PR #1, opened 2023-10-24
Incoportate opengl abstraction classes and create main player character and movement

jonmgomez:
To use opengl more nicely, abstract the frequently used features into classes
- VertexBuffer
- IndexBuffer
- VertexArray
- Shader
- Texture
- Renderer

Create a simple program using these classes and a Rectangle class to create a rectangle for the player on screen. Listen to the keyboard for 'w' and 's' keys and update position of player accordingly. Rectangle class will handle drawing the object on screen based on position


rectangle.cpp
```diff
@@ -0,0 +1,75 @@
...
+    0, 1, 2,
+    2, 3, 0
+};
+
+Rectangle::Rectangle(const glm::vec3& position, float width, float height) :
+    mPosition(position),
+    mWidth(width),
+    mHeight(height),
+    mVB(4 * 4 * sizeof(float)),
+    mVA(),
```
jonmgomez: Is there a point to even do this, if these are empty?

vertexbuffer.cpp
```diff
@@ -0,0 +1,44 @@
...
+    GLCall(glBufferData(GL_ARRAY_BUFFER, size, nullptr, GL_DYNAMIC_DRAW));
+    mSize = size;
+}
+
+VertexBuffer::~VertexBuffer()
+{
+    GLCall(glDeleteBuffers(1, &mRendererID));
+}
+
+void VertexBuffer::SetBufferData(const void* data) const
```
jonmgomez, line 28: This is interesting question. Should this be const here? It certainly can because it is not affecting the class's internals. This glBufferSubData() is populating the array with data, but on the gpu and not here. However it just seems strange to mark a function called "SetBufferData" as const. Not sure if that leads to some confusing code to read for someone later on.

main.cpp
```diff
@@ -1,10 +1,27 @@
...
+#include "rectangle.h"
+#include "renderer.h"
+#include "renderutils.h"
+#include "shader.h"
+#include "texture.h"
+#include "vertexarray.h"
+#include "vertexbuffer.h"
+#include "vertexbufferlayout.h"
+
+Rectangle* player = nullptr;
```
jonmgomez: This is a temporary thing while scope things better into more functions and classes. Player is declared here for use in the KeyCallback() and in main(), but needs to wait to be initialized after opengl is properly set up. Otherwise many issues will occur.

CMakeLists.txt
```diff
@@ -22,3 +35,9 @@ target_link_libraries(pong glm)
 # and I would like to keep /WX as well. Since I don't have an easy way to ignore the warnings
 # for the system libraries, just use /W4. Seems like /Wall has some weird extra warnings anyways.
 target_compile_options(pong PRIVATE /W4 /WX)
+
+target_compile_definitions(pong PRIVATE
+    $<$<CONFIG:Debug>:
+        DEBUG=1
```
jonmgomez, line 41: One thing I still don't understand entirely is Debug vs Release modes in cmake. Is it something specfically predefined by cmake? Is there anything else it does maybe under the hood? I got this code from somewhere else, seems like if it is debug mode it sets the DEBUG flag.

jonmgomez:
Added you here @cjappl. Definitely would like some eyes on the things I commented, as well as how I set up CMake and the libraries even if it outside the changes. Feel free to look over anything, but there is a lot of opengl internal stuff that could be skimmed over.

vertexarray.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+        GLCall(glEnableVertexAttribArray(i));
+
+// Ugly warning suppression.
+// C4312: 'type cast': conversion from 'unsigned int' to 'const void *' of greater size
+// unfortunately, this is the only way to do it. (afaik)
+#pragma warning(suppress: 4312)
+        GLCall(glVertexAttribPointer(i,
+            element.mCount, element.mType, element.mNormalized,
+            layout.GetStride(), (const void *)(offset)));
```
jonmgomez: This was a warning I was hitting. And because I want to keep it right from the start, I have /WX enabled to turn them into errors. I was not able to find a work around to doing this warning suppression. Even doing a suggested `reintepret_cast<const void *>` was also giving the same warning. Is there anything that can be done about this?

rectangle.cpp
cjappl: A couple suggestions here, hopefully this code example shows all of them:

```

// Use header file initialization, which means that your defaults are always filled in!
class Rectangle
{
private:
    float mWidth { 0.0f };
    float mHeight = 0.0f;  // equivalent (basically) to above
    std::unique_ptr<Texture *> mTexture { nullptr };
    
    SomeOtherClass {}; // Empty braces default initializes, should almost always be used if you just want your class to be initialized with all it's members
};


// all of the fields are initialized, so you don't need a default constructor, only initialize the unique things that you need for that specific constructor. In this case mTexture is nullptr!

// You can see how much boilerplate this reduces, and you can always trust your objects are initialized.

Rectangle :: Rectangle(float width) :
    mWidth { width }
    // height and texture are initialized
{
};


// Because we are using a unique ptr, we no longer need to delete the texture in the destructor
// destructor is defaulted!
// Rectangle :: ~Rectangle()
//{
    // delete texture 
//};

```

vertexarray.cpp
cjappl: I think you possibly have to do a multi step cast to ensure you aren't losing data if the type goes from 64->32 bit, something like (the extremely ugly)

```
GLCall(glVertexAttribPointer(i,
            element.mCount, element.mType, element.mNormalized,
            layout.GetStride(), reinterpret_cast<const void *>(static_cast<uintptr_t>(offset))));
```

Not sure which is better..

vertexbuffer.cpp
cjappl: If the internals of the class are updated at all you should not use const.

It appears as if the internals of the class are not updated, so this is fine to use here, but agreed if I were reading it I would kind of question it for a second. Writers choice to leave it on or off

vertexbufferlayout.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <vector>
+
+#include "renderutils.h"
+
+VertexBufferLayout::VertexBufferLayout() :
+    mStride(0)
+{
+}
+
+VertexBufferLayout::~VertexBufferLayout()
```
cjappl: Don't use empty destructors, just omit them completely :) there are some times where they can mess up your program 

Sources:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-zero
https://www.oreilly.com/library/view/c-high-performance/9781787120952/6ce5c2ae-3610-4ff1-a2d6-073df75c803d.xhtml

vertexbufferlayout.cpp
```diff
@@ -0,0 +1,51 @@
+#include "vertexbufferlayout.h"
+
+#include <vector>
+
+#include "renderutils.h"
+
+VertexBufferLayout::VertexBufferLayout() :
+    mStride(0)
```
cjappl: Rather you should initialize this and mElements in the header as described above, like:

```
class BufferLayout
{
    int mStride {0};
    std::vector<Something> mElements {};
}
```

vertexbufferlayout.cpp
```diff
@@ -0,0 +1,51 @@
...
+}
+
+template<>
+void VertexBufferLayout::Push<unsigned char>(unsigned int count)
+{
+    mElements.push_back({GL_UNSIGNED_BYTE, count, true});
+    mStride += count * sizeof(GLubyte);
+}
+
+const std::vector<VertexBufferElement>& VertexBufferLayout::GetElements() const
```
cjappl, line 34: In general you should think about not exposing your internal member variables directly to the outside world. It breaks the abstraction of your class. We did it a ton at spatial and I think it leads to a lot of issues down the road.

Instead, consider if you can do the same operations internal to the class, or just use a vector directly instead of wrapping it in a class to begin with.

vertexbufferlayout.h
```diff
@@ -0,0 +1,38 @@
...
+class VertexBufferLayout
+{
+private:
+    std::vector<VertexBufferElement> mElements;
+    unsigned int mStride;
+public:
+    VertexBufferLayout();
+    ~VertexBufferLayout();
+
+    template<typename T>
```
cjappl, line 32: HOOO BOY, WE USIN TEMPLATES NOW, GOD HELP US 😈

main.cpp
```diff
@@ -24,31 +45,101 @@ int main()
...
     std::cout << "Using OpenGL version: " << glGetString(GL_VERSION) << std::endl;
 
+    int numTextureSlots;
+    glGetIntegerv(GL_MAX_TEXTURE_IMAGE_UNITS, &numTextureSlots);
+    std::cout << "Texture Slots Available: " << numTextureSlots << std::endl;
+
+    Shader shader = Shader("D:\\code\\pong\\basic.shader");
+    Renderer renderer;
+
+    player = new Rectangle(25.0f, 225.0f);
```
cjappl: Prefer unique or shared pointer "no raw new"

```
player = make_unique<Rectangle>(25.0f, 225.0f);
```

Choose shared or unique based on how many owners it will have - 1 or many.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rsmart-smart-pointers

main.cpp
```diff
@@ -1,10 +1,27 @@
...
+
 #include <iostream>
+#include <string>
 
 #include <GL/glew.h>
 #include <GLFW/glfw3.h>
 #include <glm/glm.hpp>
 #include <glm/gtc/matrix_transform.hpp>
 
+#include "indexbuffer.h"
```
cjappl, line 3: Swap the ordering of your includes. The general ordering is recommended to be:

#include "headerFileOfThisSourceFile.h" // doesn't apply here, but in classes etc

#include "myprojheader1.h"
#include "myprojheaderN.h"

#include <TPSHeader1.h> // GL/glew etc
...

#include <stlHeader.h> // string, etc

https://softwareengineering.stackexchange.com/a/325565

indexbuffer.cpp
```diff
@@ -0,0 +1,35 @@
...
+    GLCall(glGenBuffers(1, &mRendererID));
+    GLCall(glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, mRendererID));
+    GLCall(glBufferData(GL_ELEMENT_ARRAY_BUFFER, count * sizeof(unsigned int), data, GL_STATIC_DRAW));
+    mCount = count;
+}
+
+IndexBuffer::~IndexBuffer()
+{
+    GLCall(glDeleteBuffers(1, &mRendererID));
+}
```
cjappl, line 20: If you need an explicit destructor (which it appears you do) you should define or explicitly delete the copy and move constructors. Remember the rule of 5 :)

CMakeLists.txt
cjappl: Yep you are right in your interpretation. If you want to learn more, this is called a "generator expression" and is similar to a ternary expression in C++. For example:

```
    bool isDebug = DEBUG_DEFINED ? 1 : 0;
```

Is kind of what is going on here:

If the first argument into the outer expression is 1, you replace it with the text, otherwise you replace it with zero, so here, in order you have:

```
$<$<CONFIG:Debug>:Debug=1>>

Evaluate:
$<CONFIG:Debug>

If it is true:
$<1:Debug=1>

is the first argument 1? Yes, so this statement becomes:

Debug=1;


```

https://jeremimucha.com/2021/03/cmake-generator-expressions/

cjappl (approved):
The biggest issues I would recommend changing are the use of unique pointers and class header initialization! Once you get used to them they will greatly improve your coding :)

cjappl:
If any of these suggestions aren't making sense or you want more hands on instruction on them just shout and we can zoom at some point, I know there is a lot of new stuff going on here

rectangle.cpp
jonmgomez: Yeah that is much nicer. Is it better to initalize certain types like std::string with {} or "" or just leave them be like `std::string mFilePath;`. I assumed they would be left empty anyway, but might be good for consistency?

vertexarray.cpp
jonmgomez: Yeah that managed to fix it! Definitely an incantation that line is, but I think I would like to avoid warning suppression if I can.

main.cpp
jonmgomez: No more manual refcounting for me!!!
Interestingly, when I was trying to convert the mTexture variable on Rectangle to a smart pointer, I was getting this cryptic conversion warning:
```
warning C4244: 'argument': conversion from '_Ty' to  
'unsigned char', possible loss of data
          with
          [
              _Ty=int
          ]
```
 The constructor for the SolidColorTexture I was using takes 4 unsigned chars. I was putting 255 for all, and this was causing them to be interepreted as int. Very interestingly though, depending on how I created the smart pointer was what triggered the warning
`std::unique_ptr<SolicColorTexture> temp = std::make_unique<SolidColorTexture>(255, 255, 255, 255);`
did trigger the warning
`std::unique_ptr<SolicColorTexture> temp(new SolidColorTexture(255, 255, 255, 255)`
did not trigger the warning

I did manage to fix it by just using a specified type.
```
const unsigned char kColorValue = 255;
mTexture = std::make_unique<SolidColorTexture>(kColorValue, kColorValue, kColorValue, kColorValue);
```
Seems strange

vertexbufferlayout.h
jonmgomez: WE MAKING IT OUT OF C+ WITH THIS ONE! 🗣️ 🔈

rectangle.cpp
cjappl: It's completely up to you between the first two. I would initialize all of them for sure, just for consistency. The nice thing about "braced" or "universal" initialization is that you can be consistent with all types, and in general empty braces `{}` are well known to be "empty initialization". The benefit being you do not need to provide any additional information, just the empty braces.

so:
```
float mFloat {}; // inits to zero
float mAnotherFloat {0.0f}; // totally fine too, but the first one should be obvious

MyType mSomething {}; // assumed this is just some default or empty initialization

std::string mString {}; // can assume empty string
std::string mOther {""}; // more explicit
```


I generally like the braced version, but the equals version is good too. Here is a nice article about it:

https://blog.quasar.ai/2017/03/05/cpp-braced-initialization

vertexarray.cpp
cjappl: Yeah I think you are wise to avoid erro suppression. You could break it up and add a comment if you wanted:

```
// Have to do this crazy cast because reasons
auto offsetPointer = reinterpret_cast<const void *>(static_cast<uintptr_t>(offset));
```

This is also a great opportunity to wade a little bit into appreciating `auto`. The type is already on the right side of the equals sign, why not avoid the extra clutter? :)

vertexbufferlayout.h
cjappl: 😨

main.cpp
cjappl: Very interesting... 

I think the way you did it is OK. They basically want to make sure that your number can fit into a char. Funny because you would hope the warnings would be smarter and determine the literal was <=255.

I think the way you did it is perfectly fine :) 

as for your examples, another good use of auto to save you some clutter:

```
auto temp = std::make_unique<SolidColorTexture>(255, 255, 255, 255);
```

rectangle.h
```diff
@@ -1,31 +1,37 @@
...
 #include "shader.h"
 #include "texture.h"
 #include "vertexarray.h"
 #include "vertexbuffer.h"
 
+#include <glm/glm.hpp>
+
+#include <memory>
+
+static const unsigned int kIndicies[] = {
```
cjappl, line 14: prefer std::array. Lots of benefits. You can get the c array out of it with .data() I believe

rectangle.h
```diff
@@ -1,31 +1,37 @@
...
-    VertexBuffer mVB;
-    VertexArray mVA;
-    IndexBuffer mIB;
-    Texture *mTexture;
+    float mWidth { 0.0f };
+    float mHeight { 0.0f };
+    VertexBuffer mVB { 4 * 4 * sizeof(float) };
+    VertexArray mVA {};
+    IndexBuffer mIB { kIndicies, 6 };
+    std::unique_ptr<Texture>mTexture { nullptr };
```
cjappl, line 27: lovely header initialization and unique pointer! Hope you see the usefulness soon enough :)

texture.cpp
```diff
@@ -1,16 +1,22 @@
...
 #include "stb_image.h"
 #pragma warning(pop)
 
 #include <iostream>
 
-#include "renderutils.h"
+
+Texture::Texture()
+{
+}
```
cjappl, line 16: Empty initializer can be ditched. It will just automatically be generated by the compiler I believe. Alternatively you can declare it as default in the header.

texture.cpp
```diff
@@ -76,6 +81,11 @@ SolidColorTexture::SolidColorTexture(unsigned char r, unsigned char g, unsigned
     GLCall(glBindTexture(GL_TEXTURE_2D, 0));
 }
 
+SolidColorTexture::SolidColorTexture()
+{
+}
```
cjappl, line 86: same RE: default initializer

cjappl (commented):
Good stuff. Looking forward to the next one

texture.cpp
jonmgomez: So actually I was getting some problems with my use of SolidColorTexture. However I have it defined inheriting from Texture it tries to call the default constructor and gives an error
```
error C2512: 'Texture': no appropriate default constructor available
```

Doing this `Texture("")` in the SolidColorTexture constructor alleviates it, but I don't want to call the Texture constructor in this way because it does a lot extra. (I do need to reorganize how I have these classes setup though). 
```
SolidColorTexture::SolidColorTexture(unsigned char r, unsigned char g, unsigned char b, unsigned char a)
    : Texture("")
{
```

texture.h
```diff
@@ -6,13 +6,20 @@ class Texture
...
-    int mWidth, mHeight, mBPP;
+    unsigned char* mLocalBuffer { nullptr };
 protected:
-    unsigned int mRendererID;
+    int mWidth {0};
+    int mHeight {0};
+    int mBPP {0};
+    unsigned int mRendererID {0};
 public:
+    Texture();
```
cjappl, line 16: I think here you can possibly say 

```
Texture() = default;
```

To get rid of the empty one in the source file. Not like it's a massive difference or anything

texture.h
```diff
@@ -6,13 +6,20 @@ class Texture
...
     Texture(const std::string& filePath);
```
cjappl, line 17: You should set this as "explicit" 

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c46-by-default-declare-single-argument-constructors-explicit

"C.46: By default, declare single-argument constructors explicit"

cjappl (commented):
I'll leave this review alone now, looking forward to chatting more on the next one. Shout if you have any more questions and I'll try to answer them without opening any new issues :)

PR #2, opened 2023-10-29
Allow for gameobjects to create behavior through OnStart() and OnUpdate() calls

jonmgomez:
The main thing here is abstracting away how the player and other gameobjects are handled into more like a game engine. Instead of managing their position, behavior, and rendering them from main(), it uses a system similar to Unity. The Pong class manages all object currently on the screen calling OnUpdate() on each, allowing for them to create their own behaviors. Render() is also called on them to render their current positions to the screen

Key changes here
- Renderer is now a singleton. This way the objects and mesh need no knowledge of the renderer instance or the current shader it holds. A simple Draw() call to the renderer will draw whatever is needed
- The Mesh class provides a partial interface for drawing some kind of polygon to the screen. The Rectangle and Circle classes inherit from this and create their own vertex arrays and index buffers for use with drawing. Then the Mesh's Draw() call will use these objects and give them to the Renderer for drawing.
- The GameObject class provides a partial interface for any object in the game world that needs its own behavior. It has a OnStart() function to call on instantiation in the game and OnUpdate() to be called every frame during runtime. This is a similar system to Unity.
- Input is now tracked in the static Input class. Currently it only needs a simple function to check if a key is being pressed or not. This way the Player can access this at any time for use with movement.
- And some small fixes from the previous PR

### DIFF
--- a/ball.cpp
+++ b/ball.cpp
@@ -1,0 +1,15 @@
+#include "ball.h"
+
+#include "rectangle.h"
+
+void Ball::OnStart()
+{
+    mMesh = std::make_unique<Rectangle>(25.0f, 25.0f);
+    mSpeed = 3.5f;
+}
+
+void Ball::OnUpdate()
+{
+    mPosition.x += mSpeed;
+    mPosition.y += mSpeed;
+}
--- a/ball.cpp
+++ b/ball.cpp
@@ -2,10 +2,13 @@
 
 #include "rectangle.h"
 
+static constexpr float BALL_WIDTH = 25.0f;
+static constexpr float BALL_SPEED = 3.5f;
+
 void Ball::OnStart()
 {
-    mMesh = std::make_unique<Rectangle>(25.0f, 25.0f);
-    mSpeed = 3.5f;
+    mMesh = std::make_unique<Rectangle>(BALL_WIDTH, BALL_WIDTH);
+    mSpeed = BALL_SPEED;
 }
 
 void Ball::OnUpdate()
--- a/circle.cpp
+++ b/circle.cpp
@@ -1,0 +1,59 @@
+#include "circle.h"
+
+#include <glm/glm.hpp>
+
+#define _USE_MATH_DEFINES
+#include <math.h>
+
+#include <cmath>
+#include <iostream>
+#include <numeric>
+
+Circle::Circle(float radius, int sides) :
+    mRadius {radius},
+    mSides {sides}
+{
+
+    const float circumference = 2.0f * mRadius * static_cast<float>(M_PI);
+    const float arcLength = circumference / sides;
+    const float angle = arcLength / mRadius;
+
+    float currentAngle = 0.0f;
+    float normalizedX = 1.0f;
+    float normalizedY = 0.0f;
+
+    std::vector<float> positions = {};
+    positions.reserve(sides * 4);
+    for (int i = 0; i < mSides; i++)
+    {
+        positions.push_back(normalizedX * mRadius);
+        positions.push_back(normalizedY * mRadius);
+        positions.push_back(normalizedX);
+        positions.push_back(normalizedY);
+
+        currentAngle += angle;
+        normalizedX = cos(currentAngle);
+        normalizedY = sin(currentAngle);
+    }
+
+    std::vector<unsigned int> indicies = {};
+    indicies.reserve(sides * 3);
+
+    for (int i = 0; i < sides - 2; i++)
+    {
+        indicies.push_back(0);
+        indicies.push_back(i + 1);
+        indicies.push_back(i + 2);
+    }
+
+    mVB = std::make_unique<VertexBuffer>(positions.data(), static_cast<unsigned int>(positions.size() * sizeof(float)));
+    mIB = std::make_unique<IndexBuffer>(indicies.data(), static_cast<unsigned int>(indicies.size()));
+
+    VertexBufferLayout layout;
+    layout.Push<float>(2);
+    layout.Push<float>(2);
+    mVA.AddBuffer(*mVB, layout);
+
+    const unsigned char kColorValue = 255;
+    mTexture = std::make_unique<SolidColorTexture>(kColorValue, kColorValue, kColorValue, kColorValue);
+}
--- a/circle.cpp
+++ b/circle.cpp
@@ -13,9 +13,8 @@ Circle::Circle(float radius, int sides) :
     mRadius {radius},
     mSides {sides}
 {
-
     const float circumference = 2.0f * mRadius * static_cast<float>(M_PI);
-    const float arcLength = circumference / sides;
+    const float arcLength = circumference / mSides;
     const float angle = arcLength / mRadius;
 
     float currentAngle = 0.0f;
@@ -23,7 +22,7 @@ Circle::Circle(float radius, int sides) :
     float normalizedY = 0.0f;
 
     std::vector<float> positions = {};
-    positions.reserve(sides * 4);
+    positions.reserve(mSides * 4);
     for (int i = 0; i < mSides; i++)
     {
         positions.push_back(normalizedX * mRadius);
@@ -37,22 +36,22 @@ Circle::Circle(float radius, int sides) :
     }
 
     std::vector<unsigned int> indicies = {};
-    indicies.reserve(sides * 3);
+    indicies.reserve(mSides * 3);
 
-    for (int i = 0; i < sides - 2; i++)
+    for (int i = 0; i < mSides - 2; i++)
     {
         indicies.push_back(0);
         indicies.push_back(i + 1);
         indicies.push_back(i + 2);
     }
 
-    mVB = std::make_unique<VertexBuffer>(positions.data(), static_cast<unsigned int>(positions.size() * sizeof(float)));
-    mIB = std::make_unique<IndexBuffer>(indicies.data(), static_cast<unsigned int>(indicies.size()));
+    mVB = VertexBuffer(positions.data(), static_cast<unsigned int>(positions.size() * sizeof(float)));
+    mIB = IndexBuffer(indicies.data(), static_cast<unsigned int>(indicies.size()));
 
     VertexBufferLayout layout;
     layout.Push<float>(2);
     layout.Push<float>(2);
-    mVA.AddBuffer(*mVB, layout);
+    mVA.AddBuffer(mVB, layout);
 
     const unsigned char kColorValue = 255;
     mTexture = std::make_unique<SolidColorTexture>(kColorValue, kColorValue, kColorValue, kColorValue);
--- a/gameobject.cpp
+++ b/gameobject.cpp
@@ -7,3 +7,11 @@ void GameObject::Render()
     ASSERT(mMesh != nullptr);
     mMesh->Draw(mPosition);
 }
+
+void GameObject::OnStart()
+{
+}
+
+void GameObject::OnUpdate()
+{
+}
--- a/gameobject.cpp
+++ b/gameobject.cpp
@@ -1,0 +1,9 @@
+#include "gameobject.h"
+
+#include "renderutils.h"
+
+void GameObject::Render()
+{
+    ASSERT(mMesh != nullptr);
+    mMesh->Draw(mPosition);
+}
--- a/indexbuffer.cpp
+++ b/indexbuffer.cpp
@@ -13,6 +13,42 @@ IndexBuffer::IndexBuffer(const unsigned int* data, unsigned int count)
     mCount = count;
 }
 
+IndexBuffer::IndexBuffer(const IndexBuffer& other)
+{
+    mRendererID = other.mRendererID;
+    mCount = other.mCount;
+}
+
+IndexBuffer& IndexBuffer::operator=(const IndexBuffer& other)
+{
+    if (this != &other)
+    {
+        mRendererID = other.mRendererID;
+        mCount = other.mCount;
+    }
+    return *this;
+}
+
+IndexBuffer::IndexBuffer(IndexBuffer&& other)
+{
+    mRendererID = other.mRendererID;
+    mCount = other.mCount;
+    other.mRendererID = 0;
+    other.mCount = 0;
+}
+
+IndexBuffer& IndexBuffer::operator=(IndexBuffer&& other)
+{
+    if (this != &other)
+    {
+        mRendererID = other.mRendererID;
+        mCount = other.mCount;
+        other.mRendererID = 0;
+        other.mCount = 0;
+    }
+    return *this;
+}
+
 IndexBuffer::~IndexBuffer()
 {
     GLCall(glDeleteBuffers(1, &mRendererID));
--- a/indexbuffer.cpp
+++ b/indexbuffer.cpp
@@ -6,7 +6,6 @@
 static_assert(sizeof(unsigned int) == sizeof(GLuint));
 
 IndexBuffer::IndexBuffer(const unsigned int* data, unsigned int count)
-    : mCount(0)
 {
     GLCall(glGenBuffers(1, &mRendererID));
     GLCall(glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, mRendererID));
--- a/input.cpp
+++ b/input.cpp
@@ -4,19 +4,16 @@
 
 #include <iostream>
 
-std::array<bool, 512> Input::mKeys{ false };
+// All keys are intalized to false, for not pressed
+std::array<bool, 512> Input::mKeys{};
 
 bool Input::IsKeyPressed(unsigned int keycode)
 {
     return mKeys[keycode];
 }
 
-void Input::KeyCallback(GLFWwindow* window, int key, int scancode, int action, int mods)
+void Input::KeyCallback(GLFWwindow* /*window*/, int key, int /*scancode*/, int action, int /*mods*/)
 {
-    // Suppress unused parameter warnings
-    (void)window;
-    (void)scancode;
-    (void)mods;
     std::cout << "KeyCallback() key: " << key << ", action: " << action << std::endl;
 
     // Ignore unrecognized keys
--- a/input.cpp
+++ b/input.cpp
@@ -1,0 +1,25 @@
+#include "input.h"
+
+#include <GLFW/glfw3.h>
+
+#include <iostream>
+
+std::array<bool, 512> Input::mKeys{ false };
+
+bool Input::IsKeyPressed(unsigned int keycode)
+{
+    return mKeys[keycode];
+}
+
+void Input::KeyCallback(GLFWwindow* window, int key, int scancode, int action, int mods)
+{
+    // Suppress unused parameter warnings
+    (void)window;
+    (void)scancode;
+    (void)mods;
+
+    if (action == GLFW_PRESS)
+        mKeys[key] = true;
+    else if (action == GLFW_RELEASE)
+        mKeys[key] = false;
+}
--- a/input.cpp
+++ b/input.cpp
@@ -17,6 +17,11 @@ void Input::KeyCallback(GLFWwindow* window, int key, int scancode, int action, i
     (void)window;
     (void)scancode;
     (void)mods;
+    std::cout << "KeyCallback() key: " << key << ", action: " << action << std::endl;
+
+    // Ignore unrecognized keys
+    if (key < 0)
+        return;
 
     if (action == GLFW_PRESS)
         mKeys[key] = true;
--- a/main.cpp
+++ b/main.cpp
@@ -3,6 +3,7 @@
 #include "indexbuffer.h"
 #include "input.h"
 #include "player.h"
+#include "pong.h"
 #include "rectangle.h"
 #include "renderer.h"
 #include "renderutils.h"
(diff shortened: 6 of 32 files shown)


mesh.h
```diff
@@ -0,0 +1,22 @@
...
+#include <glm/glm.hpp>
+
+#include <memory>
+
+class Mesh
+{
+protected:
+    VertexArray mVA {};
+    std::unique_ptr<VertexBuffer>mVB { nullptr };
+    std::unique_ptr<IndexBuffer>mIB { nullptr };
```
jonmgomez: Curious if you have any thoughts here @cjappl. These classes need their verticies known on instantation in order to render properly. But as a interface this Mesh class has no verticies known, so it cant really allocate these on the stack (at least with the current way these classes are setup). My best work around was making them heap allocated. I know one of our old problems was that everything in a lot of our classes were pointers, when they didn't need to be.

I think it is possible to have a late instantation, something like:
```
VertexBuffer vb;
vb.SetBufferData(positions.data(), positions.size() * sizeof(float))
```
But that is kind of a similar pattern to our old string class
```
// Create one
splString text;
// Now make it useful
text.Copy("Hello World");
```
Which I was never a fan of either

ball.cpp
```diff
@@ -0,0 +1,15 @@
+#include "ball.h"
+
+#include "rectangle.h"
+
+void Ball::OnStart()
+{
+    mMesh = std::make_unique<Rectangle>(25.0f, 25.0f);
+    mSpeed = 3.5f;
```
cjappl: Could consider making these constants somewhere, all of these magic numbers

Is it true that OnStart is called once and only once? I would half expect this to all be set up in a constructor / by default initialization in a header

circle.cpp
```diff
@@ -0,0 +1,59 @@
+#include "circle.h"
+
+#include <glm/glm.hpp>
+
+#define _USE_MATH_DEFINES
```
cjappl, line 5: I would possibly include guard this. Not sure if it will make a difference be something is making me think it:

```
#ifndef _USE_MATH_DEFINES
#define _USE_MATH_DEFINES
#endif // _USE_MATH_DEFINES

```

circle.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <math.h>
+
+#include <cmath>
```
cjappl, line 8: This should be the same as math.h, you likely do not need math.h

In general when you have a system or std library that is <something.h> it is a c-style standard header, and in C++ the recommendation is to make it <csomething>.

I wasn't entirely sure why this was the case, but chatgpt makes a compelling arugment:


In C++ why should I use the C++ standard headers instead of the C standard headers? For example, why should one prefer <cmath> over <math.h> ?

ChatGPT

Using C++ standard headers over their C counterparts in C++ code offers several advantages:

    Namespacing: The C++ headers place their contents in the std namespace, preventing potential name collisions in the global namespace. For instance, with <cmath>, functions are available as std::sqrt, std::sin, etc.

    Consistency: When writing C++ code, it's more consistent to use C++ headers across the board. This provides a uniform style and approach in your codebase.

    Overloads: C++ headers might provide additional overloads of the functions to handle C++ types or features. For instance, <cmath> might have overloads that work seamlessly with float, double, and long double.

    Templates and Types: The C++ versions may offer template functions, allowing for better type safety and possibly more optimizations during compile time.

    Error Handling: C++ headers might use exceptions or other C++ mechanisms for error handling instead of the C-style error handling mechanisms.

    Inclusion of Additional Features: Some C++ headers incorporate additional functionalities that aren't present in the C versions. For example, <cstddef> introduces types like std::size_t and std::ptrdiff_t.

    Future-Proofing: Sticking with C++ headers can make the code more resilient to potential changes in future C++ standards.

While using <math.h> in a C++ program will not result in immediate issues, it's generally a good practice to favor C++ headers when writing C++ code for the reasons mentioned above.

circle.cpp
```diff
@@ -0,0 +1,59 @@
...
+    const float circumference = 2.0f * mRadius * static_cast<float>(M_PI);
+    const float arcLength = circumference / sides;
+    const float angle = arcLength / mRadius;
+
+    float currentAngle = 0.0f;
+    float normalizedX = 1.0f;
+    float normalizedY = 0.0f;
+
+    std::vector<float> positions = {};
+    positions.reserve(sides * 4);
```
cjappl: Nice. Wanted to call this out as a very good practice. You should always consider pre-reserving if possible while using std::vector

gameobject.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+void GameObject::Render()
+{
+    ASSERT(mMesh != nullptr);
+    mMesh->Draw(mPosition);
+}
+
+void GameObject::OnStart()
+{
+}
```
cjappl, line 13: should either of these be pure virtual? Or does it make the most sense to have them both be empty in the base class? Both answers can be valid, just want you to think about it.

input.cpp
```diff
@@ -0,0 +1,30 @@
+#include "input.h"
+
+#include <GLFW/glfw3.h>
+
+#include <iostream>
+
+std::array<bool, 512> Input::mKeys{ false };
```
cjappl: For reader confusion, I would possibly omit this false. The remaining 511 entries are being default constructed (and bool default constructs with braces to be false).

I would think the slightly more understandable to an average reader would be:

```
std::array<bool, 512> Input::mKeys{ };
```

and/or leaving a comment that you know you expect this to be filled with all `false`

input.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+bool Input::IsKeyPressed(unsigned int keycode)
+{
+    return mKeys[keycode];
+}
+
+void Input::KeyCallback(GLFWwindow* window, int key, int scancode, int action, int mods)
+{
+    // Suppress unused parameter warnings
```
cjappl: The way that I would do this is comment them out of the parameter line. 

```cpp
void Input::KeyCallback(GLFWwindow* /*window*/, int key, int /*scancode*/, int action, int /*mods*/)
```

input.cpp
```diff
@@ -0,0 +1,30 @@
...
+    (void)window;
+    (void)scancode;
+    (void)mods;
+    std::cout << "KeyCallback() key: " << key << ", action: " << action << std::endl;
+
+    // Ignore unrecognized keys
+    if (key < 0)
+        return;
```
cjappl, line 21: EARLY RETURN??????????????????? ⚠️ ⚠️ you monster. I'm calling MMP

main.cpp
```diff
@@ -65,28 +55,20 @@ int main()
     glGetIntegerv(GL_MAX_TEXTURE_IMAGE_UNITS, &numTextureSlots);
     std::cout << "Texture Slots Available: " << numTextureSlots << std::endl;
 
-    Shader shader = Shader("D:\\code\\pong\\basic.shader");
-    Renderer renderer;
-
-    player = std::make_unique<Rectangle>(25.0f, 225.0f);
-    player->mPosition = glm::vec3(-550.0f, 0.0f, 0.0f);
-    const float playerSpeed = 2.5f;
+    Renderer::SetShader("D:\\code\\pong\\basic.shader");
```
cjappl, line 58: In the long run you should make this configurable via a json input file or similar, so other people can run it on other systems

mesh.h
cjappl: Ok, there are a few approaches that spring to mind here.

First off, is it possible to just copy in a new vertexBuffer when the child class is instantiated? And make a default constructor of a VertexBuffer and IndexBuffer?

That is:

```
class Mesh
{
protected:
    VertexArray mVA {};
    VertexBuffer mVB {  };
    IndexBuffer mIB {  };
```

```
    Circle::Circle(...)
    {
        mVB = VertexBuffer(positions.data(), static_cast<unsigned int>(positions.size() * sizeof(float));
    }
```

Another approach is in the Mesh class, ditching all of the protected members, and instead having private getters for them that are pure virtual:

```

    virtual  VertexArray& GetVertexArray() = 0;
    virtual  VertexBuffer& GetVertexBuffer() = 0;
    virtual IndexBuffer&   GetIndexBuffer() = 0;
```

This leaves it to the individual subclasses to populate their buffers and arrays, and no implementation details are needed in this Mesh parent class. The mesh parent class can then either operate on a pointer to a Mesh, or a template:

```
// method could be static now
void Mesh::Draw(Mesh* mesh, const glm::vec3& position) const
{
    Renderer::Draw(mesh->GetVertexArray(), mesh->GetVertexBuffer(), position, ...);
}
```

Or with templates:
```
<class MeshType>
void Mesh::Draw(const MeshType& mesh, const glm::vec3& position) const
{
    Renderer::Draw(mesh.GetVertexArray(), mesh.GetVertexBuffer(), position, ...);
}
```
You can see in this example that templates kind of act like "sameness" enforcers - similar but different to inheritance. If the type you pass in `MeshType` does not have these methods defined, the program will not compile. Just as if you didn't define the pure virtual above. In fact, in this answer you can completely remove the pure virtual methods. Just everything that needs to be passed into this "Draw" function needs to have them defined.

cjappl: The last one is so convoluted I ended up writing up a big thing and then realizing I don't understand the implications well enough to suggest it :) homework for me I guess

opponent.cpp
```diff
@@ -0,0 +1,35 @@
...
+    else
+    {
+        mPosition.y -= mSpeed;
+    }
+
+    if (mPosition.y >= 225.0f)
+    {
+        movingUp = false;
+    }
+    else if (mPosition.y <= -225.0f)
```
cjappl: Abstract away magic numbers

player.cpp
```diff
@@ -0,0 +1,28 @@
...
+    if (Input::IsKeyPressed(GLFW_KEY_W))
+    {
+        mPosition.y += mSpeed;
+    }
+    else if (Input::IsKeyPressed(GLFW_KEY_S))
+    {
+        mPosition.y -= mSpeed;
+    }
+
+    std::cout << "Position x: " << mPosition.x << ", y: " << mPosition.y << std::endl;
```
cjappl, line 32: Could consider using a logging library to get this functionality and be able to tweak it, using debug/info logging:

https://github.com/gabime/spdlog

is a typical one I see recommended.

One thing to investigate for you. If this is being called in your main game loop, you may have concerns with realtime safety, similar to in audio code. If this call takes an undetermined amount of time to come back (because it's a system call) and you need to render a frame in the same callback, you may have funky frame drops or skippy frames. I'm not sure how most game engines deal with that, or if they are as paranoid as audio devs

pong.cpp
```diff
@@ -0,0 +1,29 @@
...
+    mGameObjects.push_back(std::move(opponent));
+
+    auto ball = std::make_unique<Ball>();
+    ball->OnStart();
+    mGameObjects.push_back(std::move(ball));
+}
+
+void Pong::PongGameLoop()
+{
+    for (auto& gameObject : mGameObjects)
```
cjappl, line 24: Inspired shit 👏 

See how much nicer that is? You make it impossible to have an out of bounds error. and if the type changes you have it `auto`matically updated here. Good stuff

pong.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include "gameobject.h"
+
+#include <vector>
+
+class Pong
+{
+private:
+    std::vector<std::unique_ptr<GameObject>>mGameObjects {};
```
cjappl, line 10: Now we're std-libbing 😎 

the main downside is these heinous lengthy typenames withe a billion zillion templates. Small price to pay but still annoying.

If you ever get an especially bad one, I recommend using `using`

```
     std::vector<std::unique_ptr<std::vector<std::shared_ptr<GameObject>>mGameObjectsCollection

...

using GameObjectSharedPtr = std::shared_ptr<GameObject>;
using GameObjectSharedPtrVector = std::vector<GameObjectSharedPtr>;
using GameObjectCollection = std::unique_ptr<GameObjectSharedPtrVector>;

GameObjectCollection mGameObjectsCollection



```

rectangle.cpp
```diff
@@ -12,53 +12,30 @@
 
 #include <memory>
 
-Rectangle::Rectangle(const glm::vec3& position, float width, float height) :
-    mPosition {position},
+static const std::array<unsigned int, 6> kIndicies[] = {
```
cjappl: I think the empty braces `[]` are not needed here. You may actually be making a c array of std::arrays, and if that's the case god help us all

rectangle.cpp
```diff
@@ -12,53 +12,30 @@
...
+    0, 1, 2,
+    2, 3, 0
+};
+
+Rectangle::Rectangle(float width, float height) :
     mWidth {width},
     mHeight {height}
 {
-    float positions[] = {
+    const std::array<float, 16> positions[] = {
```
cjappl: Same, I'm not sure the `[]` is nec essary

renderutils.h
```diff
@@ -2,8 +2,14 @@
...
 #include <GL/glew.h>
 
+#include <iostream>
+
 #ifdef DEBUG
-    #define ASSERT(x) if (!(x)) __debugbreak();
+    #define ASSERT(x) if (!(x)) \
+    { \
+        std::cout << "Assertion failed: (" << #x << "), File: " __FILE__  ", Line: " << __LINE__ << std::endl; \
+        __debugbreak(); \
```
cjappl, line 11: Nice handy assert macros! Cool!

vertexarray.cpp
```diff
@@ -4,8 +4,7 @@
 
 #include <vector>
 
-VertexArray::VertexArray() :
-    mRendererID(0)
+VertexArray::VertexArray()
```
cjappl, line 7: One more high level thing!! You should be using a namespace in almost all of your custom files.

So in sources and headers, wrap everything after your includes in a:

```
#include <something>

...

namespace pong
{

VertexArray::VertexArray()
...

}
```

Why is this useful? Well, remember at spatial where everything was prefixed by `spl`? the spirit of that idea was good, but stuck in 1989. The idea is that at some point you may want to differentiate between YOUR VertexArray, and some other implementation. For instance, if you bring in some TPS that also defines a VertexArray, you may be in trouble.

Having everything in a custom namespace is considered a good practice. It allows you to group like ideas together and avoid collisions.

cjappl (approved):
Nothing glaring that I see here, just some general suggestions on things I saw

renderer.cpp
```diff
@@ -1,19 +1,43 @@
...
+#include "texture.h"
+#include "vertexarray.h"
+
+#include <glm/glm.hpp>
+#include <glm/gtc/matrix_transform.hpp>
 
 #include <iostream>
 
-void Renderer::Draw(const VertexArray& va, const IndexBuffer& ib, const Shader& shader) const
+std::unique_ptr<Renderer> Renderer::sInstance = std::make_unique<Renderer>();
```
cjappl: You could consider making this a meyer's singleton. The pattern looks like this:

```
class Singleton
{
  
public:
  static Singleton& Instance()
  {
    static Singleton s; 
    return s;
  }
  
private:
  Singleton() {}
  Singleton(const Singleton &) = delete;
  Singleton& operator= (const Singleton &) = delete;
  ~Singleton() {}

};
```

Then all access, like on line 17 looks like this:

```
instance.mShader = make_unique<Shader>...
```


Note a few benefits of this: The instance is instantiated the first time it is used, not somewhere in the insanity of static initialization. Next, you aren't dealing with pointers.

renderer.h
```diff
@@ -1,14 +1,30 @@
...
 private:
+    static std::unique_ptr<Renderer>sInstance;
+
+protected:
+    std::unique_ptr<Shader>mShader { nullptr };
+
 public:
-    void Draw(const VertexArray& va, const IndexBuffer& ib, const Shader& shader) const;
-    void Clear() const;
+    Renderer() = default;
```
cjappl, line 14: To be a true singleton, you should hide this as "private" and have the `instance` method instantiate the one true singleton. You want to enforce the fact that you will not ever have a second one of these things.

renderer.h
```diff
@@ -1,14 +1,30 @@
...
 
+#include <memory>
+
 class Renderer
 {
 private:
+    static std::unique_ptr<Renderer>sInstance;
+
+protected:
+    std::unique_ptr<Shader>mShader { nullptr };
```
cjappl: Possibly a code smell that this is protected?

circle.cpp
jonmgomez: So, interesting issue with this. I only need these headers for `M_PI`, and in order to use that I need the `#define _USE_MATH_DEFINES` too. This works fine with the `<math.h>` header, but for `<cmath>` it seems it is made in a way where if anything is included before that macro is defined it will not work properly? The only way I could get it to work was by doing 
```
#define _USE_MATH_DEFINES

#include "circle.h"

#include <cmath>
```
Where the define had to come before even the circle header file to work, so it breaks the include order a bit. Doing this in CMake also worked. `target_compile_definitions(pong PRIVATE _USE_MATH_DEFINES)`
https://stackoverflow.com/questions/6563810/m-pi-works-with-math-h-but-not-with-cmath-in-visual-studio

Leaving it for now, but curious what you think,

gameobject.cpp
jonmgomez: I figured leaving these empty in the base class is the nicest. So that if a gameobject doesn't need to implement a OnUpdate() it doesn't have to override to implement a empty one.

rectangle.cpp
jonmgomez: Yeah good catch! I think that was left over from when they were c style arrays

renderer.h
jonmgomez: Hmm I thought I was having some issues if I didn't have it as protected. Must have fixed them since then. Switched it to private

mesh.h
jonmgomez: Yeah I decided to go with your first suggestions of an empty instantation and assigning them later on. I had to implement the copy constructors to do it. If you could look at those and make sure they look good that would be great. I also implemented the move constructors just to practice a bit doing it. Thanks!

ball.cpp
jonmgomez: Yeah so I am thinking OnStart() really isn't needed. I guess I was just used to how Unity is setup that most things get instantatied in the Start() function in C#, but that really isn't need when you can just use the constructor C++. Ill leave OnStart() for now, but will probably remove it later on

input.cpp
jonmgomez: Aha I had an assumption that doing it with one value intialized them all to that value. ie `std::array<bool, 512> Input::mKeys{ true };` initialized all 512 to `true`. Not sure where I got that from.

input.cpp
jonmgomez: Neat! Didn't know you could do that.

input.cpp
jonmgomez: 😁

main.cpp
jonmgomez: Yeah that is a good idea, will get that a bit later.

pong.cpp
jonmgomez: Yeah definitely. Theres really something to all this standard library and modern c++ stuff. Where has it been all my life 🤔

renderer.cpp
jonmgomez: Yeah that is cooler. Although I couldn't access the instance just through name like you showed, I had to use `GetInstance()`. I figured that is what you meant, or that I set it up wrong.

vertexarray.cpp
jonmgomez: Interesting, yeah I will tackle that in a bit and going forward.

pong.h
jonmgomez: Yeah that looks like a real mouthful. Will keep in mind!

circle.cpp
cjappl: Very perplexing. Also hilariously these were finally added to c++ in c++20:

https://en.cppreference.com/w/cpp/numeric/constants

If you felt bold you could use c++20, I think MSVC has the best support of it of any of the major players. For the meantime though any of your solutions are totally fine, not worth stressing about

indexbuffer.cpp
```diff
@@ -6,14 +6,49 @@
...
+
+IndexBuffer::IndexBuffer(IndexBuffer&& other)
+{
+    mRendererID = other.mRendererID;
+    mCount = other.mCount;
+    other.mRendererID = 0;
+    other.mCount = 0;
+}
+
+IndexBuffer& IndexBuffer::operator=(IndexBuffer&& other)
```
cjappl, line 40: Great work!! This is great practice to get in. Seems like you've done these correctly to my knowledge :)

renderer.cpp
```diff
@@ -10,29 +10,33 @@
 
 #include <iostream>
 
-std::unique_ptr<Renderer> Renderer::sInstance = std::make_unique<Renderer>();
+Renderer& Renderer::GetInstance()
+{
+    static Renderer sInstance;
+    return sInstance;
```
cjappl, line 16: nice

cjappl (commented):
Good stuff

renderer.cpp
cjappl: Yep, my typo. You did it as I intended to write it 👍

ball.cpp
cjappl: Yeah, not the end of the world either way. You may discover one is better than the other as you get deeper in the process

PR #3, opened 2023-10-31
Add pong namespace and cleanup main()

jonmgomez:
A pretty small one. Cleanup main() a bit by splitting into functions for readability and necessity because of a program termination problem.

### DIFF
--- a/renderer.cpp
+++ b/renderer.cpp
@@ -10,6 +10,9 @@
 
 #include <iostream>
 
+namespace pong
+{
+
 Renderer& Renderer::GetInstance()
 {
     static Renderer sInstance;
@@ -45,3 +48,5 @@ void Renderer::Clear()
 {
     GLCall(glClear(GL_COLOR_BUFFER_BIT));
 }
+
+} // namespace pong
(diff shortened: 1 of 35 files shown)


renderer.cpp
```diff
@@ -10,12 +10,21 @@
...
 {
     static Renderer sInstance;
     return sInstance;
 }
 
+void Renderer::Cleanup()
+{
+    GetInstance().mShader.reset();
+    GetInstance().mShader = nullptr;
+}
```
cjappl, line 26: Is there a reason this doesn't happen in a destructor?

cjappl (approved):

renderer.cpp
jonmgomez: Yeah the issue I was having was that after closing the window, the program would not terminate until manually exited with Ctrl + C. The reason was because after calling glfwTerminate() the shader was still active since it was a static variable. In ~Shader() it cleans up in opengl using some glfw calls. But calling these after Terminate() results in some infinite loop. So they need to be manually destructed before then.

PR #4, opened 2023-10-31
Parse shader file from config and add json library

jonmgomez:
Only really need to look at main.cpp here. This is adding a json parsing library and reading in a config file form command arguments. Use the "shader" key to load the shader into the program.


CMakeLists.txt
```diff
@@ -31,13 +31,15 @@ add_executable(pong ${SOURCES})
 add_subdirectory(glfw-3.3.8)
 add_subdirectory(glew-2.1.0)
 add_subdirectory(glm-0.9.9.8)
+add_subdirectory(json-3.11.2/json-3.11.2)
```
jonmgomez: Curious what is the best way to organize tps libraries for this project @cjappl? I know we used zip files in some way before? Looking to include these in a nice way, that doesn't make this project show as being mainly an HTML repo haha

cjappl: So this is a great time to reorganize your tps. I highly recommend you zip all of these dependencies into tar.gz files to save space (and ensure you don't have some crazy 2k line PR). I would recommend you add a `tps` top level folder with it's own CMakeLists.txt

From there, the pattern is like so in your tps/CMakeLists.txt:

```
FetchContent_Declare(doctest
    URL ${CMAKE_CURRENT_SOURCE_DIR}/doctest/doctest-2.4.11.tar.gz
)
FetchContent_MakeAvailable(doctest)

```

What this does is automagically unzip/untar the thing, then `FetchContent_MakeAvailable` goes through and calls `add_subdirectory` like you are doing here

If you ever want to play around with a dependency without adding it to your repo, or you don't want to check something in, FetchContent works on git repos that are built with CMake:

```
# Fetch zlib from its Git repository
FetchContent_Declare(
    zlib
    GIT_REPOSITORY https://github.com/madler/zlib.git
    GIT_TAG v1.2.11  # You can specify a specific tag or commit hash
)
```

main.cpp
```diff
@@ -8,12 +8,38 @@
...
+        return nullptr;
+    }
+
+    nlohmann::json jsonData;
+
+    try
+    {
+        jsonFile >> jsonData;
+    }
+    catch (const nlohmann::json::parse_error& e)
```
cjappl, line 34: EXCEPTIONS??????????????????? 🤯

cjappl (approved):
Would highly recommend changing the json to be zipped up to save some sanity. :)

CMakeLists.txt
jonmgomez: Yup that is much nicer, thanks! Ill follow up with the rest of the tps libraries

main.cpp
jonmgomez: Its only been a month! I have officially lost it 🧠

main.cpp
```diff
@@ -8,12 +8,38 @@
...
 #include <glm/glm.hpp>
+#include <nlohmann/json.hpp>
 
+#include <fstream>
 #include <iostream>
 #include <string>
 
 using namespace pong;
 
+nlohmann::json OpenJsonFile(const std::string& filePath)
```
cjappl, line 19: You'll have to let me know what you think of this library. It is the big json library on the block but I've never used it myself

cjappl (commented):
Much better!!

PR #5, opened 2023-11-01
Zip other tps libs and use FetchContent to build

jonmgomez:
Zipping the rest of the tps libs into the tps folder to make things cleaner.


CMakeLists.txt
```diff
@@ -45,7 +41,7 @@ target_link_libraries(pong nlohmann_json)
 # I would use /Wall here but unfortunately the msvc headers trigger some pedantic warnings,
 # and I would like to keep /WX as well. Since I don't have an easy way to ignore the warnings
 # for the system libraries, just use /W4. Seems like /Wall has some weird extra warnings anyways.
-target_compile_options(pong PRIVATE /W4 /WX)
+target_compile_options(pong PRIVATE /W4 /WX /MP)
```
jonmgomez, line 44: Interesting MSVC quirk I found out. The -j flag does not work entirely like it works with makefiles or ninja. I noticed it would build the tps libs like glfw and glew faster, but my source files were the same build time. It seems windows wants this flag /MP on generation that will act like the -j flag and speed the rest of the build times up. Not exactly sure what causes it to be affected by one or the other

cjappl: huh! Funky.

cjappl:
This review is too big for me to take a peek at, going to assume you did it right :)

PR #6, opened 2023-11-03
Implement collisions checking and core pong game loop

jonmgomez:
GameObjects can now create a ColliderBox that will represent their phyiscial bounds to collide within the game world. The Pong class currently manages collisions, but maybed handed off to something else later. It will check to see if the objects collide. And if so will call OnCollisionStart() if these objects were not colliding in the last frame (i.e. this is the start of their collision). If they did collide last frame, call OnCollisionStay() indicating they are still colliding. And if they do not collide, but were last frame, call OnCollisionStop() to indicate they stopped colliding.

The ball will interpret collisions with the new Wall objects and the Player and Opponent to bounce off them and keep moving.

Almost playable!

### DIFF
--- a/ball.cpp
+++ b/ball.cpp
@@ -6,19 +6,35 @@ namespace pong
 {
 
 static constexpr float BALL_WIDTH = 25.0f;
-static constexpr float BALL_SPEED = 3.5f;
+static constexpr float BALL_SPEED = 4.5f;
+static int id = 0;
 
 void Ball::OnStart()
 {
     mMesh = std::make_unique<Rectangle>(BALL_WIDTH, BALL_WIDTH);
-    mSpeed = BALL_SPEED;
+    mColliderBox = std::make_unique<ColliderBox>(BALL_WIDTH, BALL_WIDTH);
+    mXSpeed = BALL_SPEED;
+    mYSpeed = BALL_SPEED;
     mName = "Ball";
 }
 
 void Ball::OnUpdate()
 {
-    mPosition.x += mSpeed;
-    mPosition.y += mSpeed;
+    const glm::vec3 newPosition = GetPosition() + glm::vec3(mXSpeed, mYSpeed, 0.0f);
+    SetPosition(newPosition);
+}
+
+void Ball::OnCollisionStart(GameObject& other)
+{
+    if (other.GetName() == "Player" || other.GetName() == "Opponent"
+     || other.GetName() == "WallV")
+    {
+        mXSpeed = -mXSpeed;
+    }
+    else
+    {
+        mYSpeed = -mYSpeed;
+    }
 }
 
 } // namespace pong
--- a/ball.cpp
+++ b/ball.cpp
@@ -12,6 +12,7 @@ void Ball::OnStart()
 {
     mMesh = std::make_unique<Rectangle>(BALL_WIDTH, BALL_WIDTH);
     mSpeed = BALL_SPEED;
+    mName = "Ball";
 }
 
 void Ball::OnUpdate()
--- a/ball.cpp
+++ b/ball.cpp
@@ -13,9 +13,9 @@ void Ball::OnStart()
 {
     mMesh = std::make_unique<Rectangle>(BALL_WIDTH, BALL_WIDTH);
     mColliderBox = std::make_unique<ColliderBox>(BALL_WIDTH, BALL_WIDTH);
+    SetInstanceName("Ball");
     mXSpeed = BALL_SPEED;
     mYSpeed = BALL_SPEED;
-    mName = "Ball";
 }
 
 void Ball::OnUpdate()
@@ -26,8 +26,8 @@ void Ball::OnUpdate()
 
 void Ball::OnCollisionStart(GameObject& other)
 {
-    if (other.GetName() == "Player" || other.GetName() == "Opponent"
-     || other.GetName() == "WallV")
+    if (other.GetInstanceName() == "Player" || other.GetInstanceName() == "Opponent"
+     || other.GetInstanceName() == "WallV")
     {
         mXSpeed = -mXSpeed;
     }
--- a/ball.h
+++ b/ball.h
@@ -8,11 +8,13 @@ namespace pong
 class Ball : public GameObject
 {
 private:
-    float mSpeed { 0.0f };
+    float mXSpeed { 0.0f };
+    float mYSpeed { 0.0f };
 
 public:
     void OnStart() override;
     void OnUpdate() override;
+    void OnCollisionStart(GameObject& other) override;
 };
 
 } // namespace pong
--- a/colliderbox.cpp
+++ b/colliderbox.cpp
@@ -2,6 +2,7 @@
 
 #include <glm/glm.hpp>
 
+#include <algorithm>
 #include <iostream>
 #include <sstream>
 
@@ -20,7 +21,7 @@ ColliderBox::ColliderBox(float width, float height)
     mPositionBounds = mBounds;
 }
 
-void ColliderBox::UpdatePositionBounds(const glm::vec3& position)
+void ColliderBox::OnPositionUpdate(const glm::vec3& position)
 {
     mPositionBounds[0] = mBounds[0] + position;
     mPositionBounds[1] = mBounds[1] + position;
@@ -33,24 +34,21 @@ bool ColliderBox::CheckForCollision(const ColliderBox& other) const
     const std::array<glm::vec3, 4>otherBounds = other.mPositionBounds;
 
     // Check within own bounds
-    for (const auto& position : otherBounds)
+
+    if (std::any_of(std::begin(otherBounds), std::end(otherBounds),
+        [&](const auto& position) { return CheckPointInBounds(position); }))
     {
-        if (CheckPointInBounds(position))
-        {
-            return true;
-        }
+        return true;
     }
 
     // Check within other collider's bounds. This is due to the
     // possibility of the other collider ecompassing this collider
     // If that occurs, the other colliders positions do not not appear
     // inside the bounds and would not register as collision
-    for (const auto& position : mPositionBounds)
+    if (std::any_of(std::begin(mPositionBounds), std::end(mPositionBounds),
+        [&other](const auto& position) { return other.CheckPointInBounds(position); }))
     {
-        if (other.CheckPointInBounds(position))
-        {
-            return true;
-        }
+        return true;
     }
 
     return false;
--- a/colliderbox.cpp
+++ b/colliderbox.cpp
@@ -1,0 +1,65 @@
+#include "colliderbox.h"
+
+#include <glm/glm.hpp>
+
+#include <iostream>
+#include <sstream>
+
+namespace pong
+{
+
+ColliderBox::ColliderBox(float width, float height)
+{
+    mBounds = {
+        glm::vec3(-width / 2.0f, -height / 2.0f, 0.0f),
+        glm::vec3( width / 2.0f,  height / 2.0f, 0.0f),
+        glm::vec3(-width / 2.0f,  height / 2.0f, 0.0f),
+        glm::vec3( width / 2.0f, -height / 2.0f, 0.0f)
+    };
+
+    mPositionBounds = mBounds;
+}
+
+void ColliderBox::UpdatePositionBounds(const glm::vec3& position)
+{
+    mPositionBounds[0] = mBounds[0] + position;
+    mPositionBounds[1] = mBounds[1] + position;
+    mPositionBounds[2] = mBounds[2] + position;
+    mPositionBounds[3] = mBounds[3] + position;
+}
+
+bool ColliderBox::CheckForCollision(const ColliderBox& other) const
+{
+    const std::array<glm::vec3, 4>otherBounds = other.mPositionBounds;
+
+    // Check within own bounds
+    for (const auto& position : otherBounds)
+    {
+        if (CheckPointInBounds(position))
+        {
+            return true;
+        }
+    }
+
+    // Check within other collider's bounds. This is due to the
+    // possibility of the other collider ecompassing this collider
+    // If that occurs, the other colliders positions do not not appear
+    // inside the bounds and would not register as collision
+    for (const auto& position : mPositionBounds)
+    {
+        if (other.CheckPointInBounds(position))
+        {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+bool ColliderBox::CheckPointInBounds(const glm::vec3& position) const
+{
+    return position.x >= mPositionBounds[0].x && position.x <= mPositionBounds[1].x
+        && position.y >= mPositionBounds[0].y && position.y <= mPositionBounds[2].y;
+}
+
+} // namespace pong
--- a/collisionmanager.cpp
+++ b/collisionmanager.cpp
@@ -2,6 +2,7 @@
 
 #include "gameobject.h"
 
+#include <algorithm>
 #include <memory>
 #include <vector>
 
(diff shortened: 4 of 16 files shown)


player.cpp
```diff
@@ -17,22 +17,22 @@ static constexpr glm::vec3 PLAYER_POSITION(-550.0f, 0.0f, 0.0f);
...
-        mPosition.y += mSpeed;
+        SetPosition(GetPosition() + glm::vec3(0.0f, mSpeed, 0.0f));
     }
     else if (Input::IsKeyPressed(GLFW_KEY_S))
     {
-        mPosition.y -= mSpeed;
+        SetPosition(GetPosition() + glm::vec3(0.0f, -mSpeed, 0.0f));
     }
-
-    std::cout << "Position x: " << mPosition.x << ", y: " << mPosition.y << std::endl;
```
jonmgomez, line 35: An update to your previous comment about printing in the game loop @cjappl . Something will definitely be needed to make this safer. Testing some of the collisions stuff, I was printing out a lot of debugging information. At least 10-15 different print statements per game loop. This caused the FPS to drop to ~75 frames. Curious, I removed all print statements and made the FPS cap unlimited. I measured an average FPS of ~4200. Even this single position print statement here in the game loop will drop the FPS from ~4200 to ~1700, haha. Perhaps a real-time logger is needed???? 😱

colliderbox.cpp
```diff
@@ -0,0 +1,65 @@
...
+        glm::vec3(-width / 2.0f, -height / 2.0f, 0.0f),
+        glm::vec3( width / 2.0f,  height / 2.0f, 0.0f),
+        glm::vec3(-width / 2.0f,  height / 2.0f, 0.0f),
+        glm::vec3( width / 2.0f, -height / 2.0f, 0.0f)
+    };
+
+    mPositionBounds = mBounds;
+}
+
+void ColliderBox::UpdatePositionBounds(const glm::vec3& position)
```
jonmgomez: I am curious if there is a better way around this. This function should be called whenever the gameobject it is attached to updates its own position, so it will update here. However to call it from the gameobject it is public and be called anywhere. But realistically this should never be called from anywhere else. Is there is a good way to solve that?

player.cpp
cjappl: If only you knew how to implement one?  ;) 

https://www.youtube.com/watch?v=4KFFMGTQIFM

https://github.com/cjappl/rtlog-cpp

I am very curious if all game rendering loops are necessary to be realtime safe. It certainly has all the smells of needing to be realtime safe. You have a hard deadline that if you don't hit it you have consequences (frame drops). I wonder how much game programmers obsess about this stuff like the audio psychos. I may have to investigate a bit, maybe I have something to learn.

ball.cpp
```diff
@@ -6,18 +6,35 @@ namespace pong
 {
 
 static constexpr float BALL_WIDTH = 25.0f;
-static constexpr float BALL_SPEED = 3.5f;
+static constexpr float BALL_SPEED = 4.5f;
+static int id = 0;
```
cjappl, line 10: Is this ever incremented from multiple threads? if so, `atomic`

ball.cpp
```diff
@@ -6,18 +6,35 @@ namespace pong
...
-    mPosition.x += mSpeed;
-    mPosition.y += mSpeed;
+    const glm::vec3 newPosition = GetPosition() + glm::vec3(mXSpeed, mYSpeed, 0.0f);
+    SetPosition(newPosition);
+}
+
+void Ball::OnCollisionStart(GameObject& other)
+{
+    if (other.GetName() == "Player" || other.GetName() == "Opponent"
+     || other.GetName() == "WallV")
```
jonmgomez: Temporary (maybe?) thing to bounce the ball in the correct direction based on the objects set name. In unity I would normally raycast from the object's previous position to the object it collided with. Then get the normal vector from the raycast and bounce the object that way. But that seems perhaps a little too advanced for right now 😬

ball.cpp
```diff
@@ -6,18 +6,35 @@ namespace pong
...
 
 void Ball::OnUpdate()
 {
-    mPosition.x += mSpeed;
-    mPosition.y += mSpeed;
+    const glm::vec3 newPosition = GetPosition() + glm::vec3(mXSpeed, mYSpeed, 0.0f);
+    SetPosition(newPosition);
+}
+
+void Ball::OnCollisionStart(GameObject& other)
```
cjappl, line 27: I think it does make sense to have a "collision manager" possibly. Some good advice is that "Classes should only have one single responsibility" this is known as the Single Responsibility Principle from the SOLID design principles. 

A Ball having to exist in space is one responsibility, having to deal with collisions is another (and hell, even drawing is a third).

You may find as you continue designing that things may get unwieldy, so keep an eye on it and refactor if you run into some pain.

ball.cpp
cjappl: At some point it may make sense to give objects enumerated Ids. This string compare can take a little while (O(n), where N is the length of the string). You could imagine a distinct set of actors/objects in your world that could be nicely condensed into an enum, which would make all of these easily comparable in O(1)

ball.h
```diff
@@ -8,11 +8,13 @@ namespace pong
...
 {
 private:
-    float mSpeed { 0.0f };
+    float mXSpeed { 0.0f };
+    float mYSpeed { 0.0f };
 
 public:
     void OnStart() override;
     void OnUpdate() override;
+    void OnCollisionStart(GameObject& other) override;
```
cjappl, line 17: Other should be const ref here, I think

colliderbox.cpp
cjappl: A couple thoughts.

This is what we typically think of as an event handler. aka it happens when some event happens, and we want a response. This implies the name should be prepended with "On". This implies to me, the idiot reader, that this funciton is called when something else happens.

`OnPositionUpdate` could be a good one.

colliderbox.cpp
```diff
@@ -0,0 +1,65 @@
...
+        {
+            return true;
+        }
+    }
+
+    // Check within other collider's bounds. This is due to the
+    // possibility of the other collider ecompassing this collider
+    // If that occurs, the other colliders positions do not not appear
+    // inside the bounds and would not register as collision
+    for (const auto& position : mPositionBounds)
```
cjappl: Good opportunity to consider playing with a standard algorithm for fun:

```
return std::any_of(std::begin(mPositionBounds), std::end(mPositionBounds), 
    [&other](const auto& position) { return other.CheckPointInBounds(position); });
```

player.cpp
jonmgomez: I imagine it is kind of somewhere in the middle. It would make sense that the game engine would be as realtime safe as possible. But at the same time, for example, Unity exposes the Instantiate() call to create a new GameObject on the fly. This certainly has to allocate somewhere and I had trouble before with this call specifically dropping the FPS because it can be expensive. To fix it, I had to use an object pool to instantiate a bunch of objects at the start, and reuse them later on. So I guess maybe it is more relaxed until it needs to be optimized better? And it seems like Instantiate() is pretty necessary across Unity too. So I don't think there is anyway around it, at least if an object pool is not overkill.

gameobject.cpp
```diff
@@ -19,4 +15,66 @@ void GameObject::OnUpdate()
...
+glm::vec3 GameObject::GetPosition() const
+{
+    return mPosition;
+}
+
+void GameObject::SetPosition(const glm::vec3& position)
+{
+    mPosition = position;
+
+    if (mColliderBox != nullptr)
```
cjappl, line 49: Is this a valid state for the object to be in? If not, let it crash here and it acts as an assert

gameobject.cpp
```diff
@@ -19,4 +15,66 @@ void GameObject::OnUpdate()
...
+    {
+        mColliderBox->UpdatePositionBounds(mPosition);
+    }
+}
+
+const std::string& GameObject::GetName() const
```
cjappl: Consider just returning the string by value (copying). I doubt this is a perf bottleneck and if someone holds on to this reference longer than the existence of the object you may be in pain.

gameobject.cpp
```diff
@@ -19,4 +15,66 @@ void GameObject::OnUpdate()
...
+void GameObject::SetName(const std::string& name)
+{
+    mName = name;
+}
+
+bool GameObject::CheckForCollision(GameObject& other)
+{
+    if (mColliderBox != nullptr && other.mColliderBox != nullptr)
+    {
+        return mColliderBox->CheckForCollision(*other.mColliderBox);
```
cjappl, line 69: same thing here, if this isn't a valid state, just let it crash

main.cpp
```diff
@@ -63,7 +63,7 @@ GLFWwindow* SetupGLFW()
     glfwMakeContextCurrent(window);
     glfwSetKeyCallback(window, Input::KeyCallback);
 
-    // Uncomment this if you would like to blow up the GPU :D
+    //Uncomment this if you would like to blow up the GPU :D
```
cjappl, line 66: 🤣

opponent.cpp
```diff
@@ -28,18 +30,18 @@ void Opponent::OnUpdate()
 {
     if (movingUp)
     {
-        mPosition.y += mSpeed;
+        SetPosition(GetPosition() + glm::vec3(0.0f, mSpeed, 0.0f));
```
cjappl: lots of use of GetPosition here, could pull it out into a variable

pong.cpp
```diff
@@ -20,15 +22,110 @@ void Pong::PongInit()
     auto ball = std::make_unique<Ball>();
     ball->OnStart();
     mGameObjects.push_back(std::move(ball));
+
+    const float horizontalWallWidth = 640 * 2;
+    const float horizontalWallHeight = 25;
+    const float horizontalWallY = 480;
+    const float verticalWallWidth = 25;
+    const float verticalWallHeight = 480 * 2;
+    const float verticalWallX = 640;
```
cjappl: all of the above (especially the multiplications), could be constexpr and calculated at compile time

pong.cpp
```diff
@@ -20,15 +22,110 @@ void Pong::PongInit()
...
+
+    auto topWall = std::make_unique<Wall>(horizontalWallWidth, horizontalWallHeight);
+    topWall->OnStart();
+    topWall->SetPosition(glm::vec3(0.0f, horizontalWallY, 0.0f));
```
cjappl, line 35: So this pattern repeats itself a handful of times. Does it make sense to have an "Setup" method or similar to a game object that calls OnStart, then positions in it's starting position? Saves yourself from some repetition. You could also consider OnStart setting the initial position

If you wanted to be very extra you could make a factory method for these which would do the pattern:

```
obj = make_unique(args)
obj.SetUp()
```

pong.cpp
```diff
@@ -20,15 +22,110 @@ void Pong::PongInit()
...
+bool Pong::IsCurrentlyColliding(int firstGameObjectId, int secondGameObjectId) const
+{
+    for (auto& collisionPair : mCurrentCollisions)
+    {
+        if (IsCollisionPairValid(collisionPair, firstGameObjectId, secondGameObjectId))
+        {
+            return true;
+        }
+    }
+    return false;
```
cjappl: another possibility for any_of.

If you ever want a really killer talk to get you acquainted with the `<algorithm>` header, I recommend:

https://www.youtube.com/watch?v=2olsGf6JIkU

pong.cpp
```diff
@@ -20,15 +22,110 @@ void Pong::PongInit()
...
+
+bool Pong::IsCollisionPairValid(const CollisionPair& pair, int firstGameObjectId, int secondGameObjectId) const
+{
+    return pair.mFirstGameObjectId == firstGameObjectId && pair.mSecondGameObjectId == secondGameObjectId
+        || pair.mFirstGameObjectId == secondGameObjectId && pair.mSecondGameObjectId == firstGameObjectId;
+}
+
+void Pong::RemoveGameObjectCollisionPair(int firstGameObjectId, int secondGameObjectId)
+{
+    mCurrentCollisions.erase(std::remove_if(mCurrentCollisions.begin(), mCurrentCollisions.end(),
```
cjappl: now you're C++ing, the old erase-remove idiom

gameobject.h
```diff
@@ -11,18 +12,37 @@ namespace pong
...
+private:
+    static int sId;
+
+    int mId { sId++ };
+
 protected:
     std::unique_ptr<Mesh>mMesh { nullptr };
+    std::unique_ptr<ColliderBox>mColliderBox { nullptr };
+    glm::vec3 mPosition { glm::vec3(0.0f) };
+    std::string mName { "" };
```
cjappl: I would encourage you to just make a GetName() method as a pure virtual, enabling the subclasses to implement these as they need (for instance wall can just return `"Wall"` without messing with mName.

In general, protected data members are a code smell of too-tight coupling between superclasses and subclasses, and should be avoided where ever possible. You may find it to be a maintenance nightmare later on

colliderbox.cpp
cjappl: In terms of preventing it being called from anywhere else, I don't think there is a decent way to do it.

The technical way to do it would be making this private, then making the two classes `friend`s, but I think that is more evil than good.

cjappl (approved):

player.cpp
cjappl: Nice. Yeah object pool sounds like a killer solution to this

ball.cpp
jonmgomez: Not for now at least. Everything is single threaded so far, but that might change in the future

ball.h
jonmgomez: I think it is nicer to let it be non-const. My reasoning is that sometimes it is nicer to do both operations on a collision between gameobjects from one side of the collision. For example a collision between some player/projectile would call TakeDamage() on the player and Destroy() on the projectile. Instead of calling each from their own respective OnCollisionStart() functions, the player only needs to implement the collision function and can call projectile.Destroy() as well. Sometimes it is just quicker to know what the behavior of a collision between two objects should be.

gameobject.cpp
jonmgomez: Yeah it is valid. The idea is that a gameobject may not want collisions or to even be rendered to the screen, so their mMesh and mColliderBoxes would be null. Like some kind of controller class for keeping track of score or ui elements would not really want collisions.

main.cpp
jonmgomez: Hah yeah. It is cool to uncomment this because it will uncap the fps limit. Normally opengl is tied to whatever the refresh rate of the monitor is. Whenever I uncap it though my GPU usage instantly jumps to and stays at 100% lol

pong.cpp
jonmgomez: Maybe. I think it might be alright for now. I have been thinking about how this kind of setup would work. Modelling things from unity, you would place all your gameobjects in the world in the 3D viewer for however you want the game to start. You would set their inital positions, names, and even internal variables you chose to expose like speed. So my thinking would be that there are some class specific setup that would be done here that could not really be put into a function.

gameobject.h
jonmgomez: I guess my issue here is that I needed mName to be changeable between instances of the same gameobject. This is just for use with the ball, so it knows which direction to bounce based on which Wall instance was hit.

colliderbox.cpp
jonmgomez: Man, it is reminding me more and more of JavaScript. Of course with much more syntax! Neat! 😁

gameobject.h
cjappl: in that case I would possibly recommend a pure virtual getName/setName at some point in the hierarchy instead of a protected member, but either way gets the job done

collisionmanager.cpp
```diff
@@ -0,0 +1,68 @@
...
+                gameObject->OnCollisionStart(*otherGameObject);
+                otherGameObject->OnCollisionStart(*gameObject);
+                mCurrentCollisions.push_back({ gameObject->GetId(), otherGameObject->GetId() });
+            }
+            else if (collision)
+            {
+                gameObject->OnCollisionStay(*otherGameObject);
+                otherGameObject->OnCollisionStay(*gameObject);
+            }
+            else if (wereColliding)
```
cjappl, line 35: This kind of branching structure possibly indicates the need for a state machine. Any time you have these if/else branches that become more complex, covering cases where things are going in and out of conditionals (never had a collision, new collision, collision continued, collision ending) you should consider a state machine. As it goes along, the maintenance burden of the state machine becomes much less than maintaining all branches of a statement like this

cjappl (commented):
Good stuff, see you on the next one 👍

PR #7, opened 2023-11-07
Add scoring, better paddle collisions, logging library. 

jonmgomez:
Quite a few things in here:
- ScoreAreas now exist behind each paddle. When the ball collides with them they will increment the score using a reference to the ScoreController. GameObjects can be found by type during runtime.
- Ball now bounces at different angle based on where it hit the paddle. Ball also speeds up on each collision.
- Opponent now has some basic AI to follow the balls current location.
- Add the spdlog library for logging output.
- Fix paddles from moving through walls.
- Some other changes/improvements.


ball.cpp
```diff
@@ -1,40 +1,80 @@
...
+        const float yPosDiff = GetPosition().y - other.GetPosition().y;
+        const float paddleHeight = other.GetColliderBox()->GetHeight();
+        const float pointInHeight = paddleHeight / 2 - abs(yPosDiff);
+
+        float percent = abs(pointInHeight / (paddleHeight / 2));
+        percent = std::max(0.35f, percent);
+        percent = std::min(1.0f, percent);
+
+        float xDir = percent;
+        float yDir = (1 - percent);
```
cjappl: would recommend using 1.0f here, just to remain in the realm of floats

ball.cpp
```diff
@@ -1,40 +1,80 @@
...
+    }
+    else if (other.GetInstanceName() == "Player" || other.GetInstanceName() == "Opponent")
+    {
+        mSpeed += BALL_SPEED_BOUNCE_INCREMENT;
+
+        const float yPosDiff = GetPosition().y - other.GetPosition().y;
+        const float paddleHeight = other.GetColliderBox()->GetHeight();
+        const float pointInHeight = paddleHeight / 2 - abs(yPosDiff);
+
+        float percent = abs(pointInHeight / (paddleHeight / 2));
```
cjappl: could consider absf, most math things have a double version and a float version

aka

sin works on doubles
sinf works on floats

gameobject.h
```diff
@@ -43,6 +43,11 @@ class GameObject
     bool CheckForCollision(GameObject& other);
 
     void Render() const;
+    template <typename T>
+    T* GetObject()
```
jonmgomez, line 47: This function is meant to get the actual derived gameobject class from an object. In general it would be something like
```
if (object->GetInstanceName("Ball"))
{
     Ball* ball = object->GetObject<Ball>();
     ball->SomeFunction();
}
```
or 
```
if (object->GetObject<Ball>() != nullptr)
{
    object->GetObject<Ball>()->SomeFunction();
}
```
to use the actual class in some way if needed, like in OnCollisionStart() for example. I know raw pointers are advised against, but I can't really return a reference here either? Becuase I need to be able to check if it is null if the object is not of this type. But I don't think I could return a shared_ptr here either, becuase that would conflict the shared_ptr in the Pong class' list of gameobject. Then it would cause a double release? What do you think? Is there a better way to do this in general?

pong.h
```diff
@@ -12,12 +12,36 @@ namespace pong
...
 private:
-    std::vector<std::unique_ptr<GameObject>>mGameObjects {};
+    Pong() = default;
+    Pong(const Pong&) = delete;
+    Pong& operator=(const Pong&) = delete;
+    Pong(Pong&&) = delete;
+    Pong& operator=(Pong&&) = delete;
+    ~Pong() = default;
+
+    std::vector<std::shared_ptr<GameObject>>mGameObjects {};
```
jonmgomez: Very curious what you think of this. I changed this to a shared_ptr array instead of unique_ptr. Realisitically I think GameObjects really are only useful when they are being using in the game itself. i.e. actually exists in this vector and have Update() and Render() called on them. So this class is generally going to be the sole "owner" of them. Other gameobjects can certainly have references to gameobjects. But, I want something similar to Unity's Destroy() api. This removes a gameobject during runtime. After this is called, the GameObject should probably not be used anymore, so I am curious if shared_ptr is better than unique_ptr?

Secondary question, if I do use unique_ptr, if another gameobject needs a reference to the same gameobject it would likely be using a reference&? I figured raw pointers are avoided, but unique_ptr could also not be used for the type, because that would cause memory to be released twice?

pong.h
```diff
@@ -12,12 +12,36 @@ namespace pong
...
-    void PongInit();
-    void PongGameLoop();
+    static Pong& GetInstance();
+
+    static void Init();
+    static void GameLoop();
+    static void Cleanup();
+
+    template<typename T>
+    static std::shared_ptr<T> FindGameObject()
```
jonmgomez: The only way I was able to make this function work was by implementing it fully in the header file. I am guessing it is some weird template issue?

scorearea.cpp
```diff
@@ -0,0 +1,51 @@
...
+            spdlog::info("Player scored!");
+            mScoreController->PlayerScored();
+        }
+        else
+        {
+            spdlog::info("Opponent scored!");
+            mScoreController->OpponentScored();
+        }
+
+        spdlog::info("New Score: P {} - O {}", mScoreController->GetPlayerScore(), mScoreController->GetOpponentScore());
```
jonmgomez, line 47: So I definitely want to get a real time logger in, but spdlog has no support for sequence numbers it seems. How would I be able to insert sequence numbers into these logs? I know we used macros for our logging, and I can imagine a way to do that by inserting an sequence number into the log call itself. Curious if you had anything off the top of your head.

ball.cpp
```diff
@@ -1,40 +1,80 @@
...
+        mSpeed += BALL_SPEED_BOUNCE_INCREMENT;
+        mVelocity = glm::normalize(glm::vec3(mVelocity.x, -mVelocity.y, 0.0f)) * mSpeed;
     }
 }
 
+void Ball::ResetBall()
+{
+    mSpeed = BALL_START_SPEED;
+
+    const float yStartingPos = rand() % (static_cast<int>(Y_STARTING_POSITION_BOUNDS) * 2) - Y_STARTING_POSITION_BOUNDS;
```
cjappl: Don't use rand as your random number generator, use one of the fancier ones, this stack overflow may get you started:

https://stackoverflow.com/questions/19665818/generate-random-numbers-using-c11-random-library

In your context, it may not matter as much to have "high quality random numbers", but it's good practice to use one of the other number generators, as using rand() in production is generally disallowed at many companies

collisionmanager.cpp
```diff
@@ -9,7 +9,7 @@
 namespace pong
 {
 
-void CollisionManager::ProcessCollisions(const std::vector<std::unique_ptr<GameObject>>& gameObjects)
+void CollisionManager::ProcessCollisions(const std::vector<std::shared_ptr<GameObject>>& gameObjects)
```
cjappl: Can you justify your switching to shared ptr for me? Why does it make sense for these objects to have shared ownership instead of single ownership?

(not critiquing you, just testing your knowledge and seeing what thought you've put into it)

gameobject.h
```diff
@@ -43,6 +43,11 @@ class GameObject
...
+    {
+        return dynamic_cast<T*>(this);
```
cjappl, line 49: This function leaves me feeling a bit funny. First off, because it's at compile time I think you could possibly get away with static_cast. Second, in what situation would you need to get an object in another type?

main.cpp
```diff
@@ -55,39 +56,47 @@ GLFWwindow* SetupGLFW()
...
     GLCall(glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA));
 
-    std::cout << "Using OpenGL version: " << glGetString(GL_VERSION) << std::endl;
+    std::string glVersion = reinterpret_cast<const char*>(glGetString(GL_VERSION));
+    spdlog::info("Using OpenGL version: {}", glVersion);
 
     int numTextureSlots;
     glGetIntegerv(GL_MAX_TEXTURE_IMAGE_UNITS, &numTextureSlots);
-    std::cout << "Texture Slots Available: " << numTextureSlots << std::endl;
+    spdlog::info("Texture Slots Available: {}", numTextureSlots);
```
cjappl, line 92: fancy fstrings!! What a time to be  alive

main.cpp
```diff
@@ -55,39 +56,47 @@ GLFWwindow* SetupGLFW()
...
-        std::cerr << "glewInit() failure" << std::endl;
+        spdlog::error("glewInit() failure");
         return nullptr;
     }
 
     GLCall(glEnable(GL_BLEND));
     GLCall(glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA));
 
-    std::cout << "Using OpenGL version: " << glGetString(GL_VERSION) << std::endl;
+    std::string glVersion = reinterpret_cast<const char*>(glGetString(GL_VERSION));
```
cjappl: the input type here should be `const char *`, alternatively you could make this auto and save these little weird conversions


```
const char* glVersion = reinterpret_cast<const char*>(glGetString(GL_VERSION));

auto glVersion = reinterpret_cast<const char*>(glGetString(GL_VERSION));
```

Basically you're superfluously building a std::string here, which seems minorly excessing

opponent.h
```diff
@@ -8,11 +9,14 @@ namespace pong
 class Opponent : public GameObject
 {
 private:
-    float mSpeed { 0.0f };
+    glm::vec3 mVelocity { 0.0f };
+    std::shared_ptr<Ball> mBall { nullptr };
```
cjappl: This seems incorrect. Does the opponent OWN the ball? I think it likely just has a passing interest in it. I would say make this a plain pointer or plain ref.

pong.h
cjappl: This is going to be relatively costly at runtime to do, dynamic casts are very expensive.

The [google coding guidelines](https://google.github.io/styleguide/cppguide.html) advise against dynamic_cast and "Run Time Type Inference" (RTTI): 

RTTI has legitimate uses but is prone to abuse, so you must be careful when using it. You may use it freely in unittests, but avoid it when possible in other code. In particular, think twice before using RTTI in new code. If you find yourself needing to write code that behaves differently based on the class of an object, consider one of the following alternatives to querying the type:

    Virtual methods are the preferred way of executing different code paths depending on a specific subclass type. This puts the work within the object itself.
    If the work belongs outside the object and instead in some processing code, consider a double-dispatch solution, such as the Visitor design pattern. This allows a facility outside the object itself to determine the type of class using the built-in type system.

cjappl (approved):

gameobject.h
jonmgomez: With functions like OnCollisionStart() you get a GameObject&. However if when handling a collision you need something specific from the other object you would need to get it into that type first. A lot of this comes from how unity handles it. For example, in Unity this is a typical instance of handling a collision.
```
public void OnTriggerEnter(Collider collider)
{
    if (collider.tag == "projectile")
    {
        // Get the actual object out from the collider.
        Projectile projectile = collider.GetComponent<Projectile>();
        // could check here to make sure it actually was an instance of Projectile (projectile != null)
        TakeDamage(projectile.GetDamage());
     }
}

File: opponent.h (hunk 12 above)
Comment (jonmgomez) [reply opponent.h]: Yeah I guess that makes more sense. Although is either pointer or ref better? I feel I have heard of "no raw pointers" which instictually makes me use smart pointers instead. It seems like this type of situation complicates it though.

File: collisionmanager.cpp (hunk 8 above)
Comment (jonmgomez) [reply collisionmanager.cpp]: Check out my one of my comments I put here! I was curious of your thoughts on this very thing.

File: opponent.h (hunk 12 above)
Comment (cjappl) [reply opponent.h]: Yep, you're close to the right saying! 

The actual saying is "no raw new" meaning you should never see just the use of `new` in a codebase, indicating they should use make_unique or similar to do the same thing. (There is also "no raw loops" which means don't use a for-loop if there is a std::algorithm that works, like `std::any_of` that we used in an earlier review).


In general, here is how you should think about the 3 major pointer types. They are all fundamentally in terms of "ownership". When you use a pointer, think about where in your abstraction the thing is OWNED.

Unique pointer:
- One unique owner
- If you hold the unique pointer, you are that objects owner

Shared pointer:
- Multiple owners
- If you hold a shared pointer, you are one of multiple owners
- too many owners of a shared pointer is a code smell

Raw pointer (or reference):
- Non-owning reference to an object
- If you hold a raw pointer, or reference, you are likely just borrowing it. You have no ownership responsibilities

Sources:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-owner
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-ptr
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r4-a-raw-reference-a-t-is-non-owning

File: collisionmanager.cpp (hunk 8 above)
Comment (cjappl) [reply collisionmanager.cpp]: I am missing the comment, I think, could you link me or give me a line number?

Comment (jonmgomez) [reply collisionmanager.cpp]: Yeah I left a few, weid that it is not showing up.  Here is that question https://github.com/jonmgomez/pong/pull/7#discussion_r1385685102

and the other comments
https://github.com/jonmgomez/pong/pull/7#discussion_r1385685057
https://github.com/jonmgomez/pong/pull/7#discussion_r1385685128
https://github.com/jonmgomez/pong/pull/7#discussion_r1385685159
let me know if it still gives problems, I can recomment them if need be

File: pong.h (hunk 5 above)
Comment (jonmgomez) [reply pong.h]: Well in that case what is a better way to do this? This function is mainly to find specific gameobjects during runtime. Like with the Opponent it needs a reference to the ball gameobject so finds it with this function in OnStart(). There certainly could be something like like `FindGameObject("Ball")` keying on its name, but I feel that has its own problems. 

Basing things off of Unity (suprise, suprise) this is a similar interface
```
public void Start()
{
     mPlayer = FindObjectOfType<Player>()

     // Or for many objects
     List<Enemy> enemies = FindObjectsOfType<Enemy>();
}
```

File: collisionmanager.cpp (hunk 8 above)
Comment (cjappl) [reply collisionmanager.cpp]: Got them now!! Thanks for the links

File: gameobject.h (hunk 3 above)
Comment (cjappl) [reply gameobject.h:47]: Raw pointers are totally good in the right circumstance :) As described in another comment, just avoid "raw new". I think the guidance is "use references whenever you can, and use pointers when you must". Thist may be one of those cases!!

In general, the options for returning a "Something or nothing" value in C++ is nullptr, so I think that's totally good. In the future, I think it will move more towards using std::optional to allow for that with value semantics.

> But I don't think I could return a shared_ptr here either, becuase that would conflict the shared_ptr in the Pong class' list of gameobject.

Yep, you shouldn't ever wrap something you didn't allocate manually in a shared pointer. You have good intuition there.

I think given what you've said about Unity doing it this way, this is fine for now! Have you peeked under the hood at Unity and confirmed they do it this way? Wonder if they have some kind of other tricky stuff up their sleeve

File: pong.h (hunk 4 above)
Comment (cjappl) [reply pong.h]: > So this class is generally going to be the sole "owner" of them.

This is the strongest case for it to be unique pointer

> Other gameobjects can certainly have references to gameobjects.

To hand out a non-owning handle to an object, give out a raw pointer or raw reference to it with

```
doSomething(myPointer.get())

doSomethingRef(*myPointer.get()); // Note this could copy it if a calling function doesn't pass by reference
```

File: pong.h (hunk 5 above)
Comment (cjappl) [reply pong.h]: yep correct! Templates need to live in the header

Comment (cjappl) [reply pong.h]: interesting. I would say spend a second looking into unity and seeing if you can determine what they are doing under the hood. If there is prior art for this, let 'r' rip :)

File: scorearea.cpp (hunk 6 above)
Comment (cjappl) [reply scorearea.cpp:47]: Yeah, a macro seems smart I think, just automatically extending the log line to put in a sequence number as it increments.

Alternatively look into making some constexpr function for it. Like macros, but type safe and may not bite you with some weirdness. They have some funny quirks that are good to learn.

**Review by cjappl — APPROVED** [review]

File: pong.h (hunk 5 above)
Comment (jonmgomez) [reply pong.h]: Yeah, it is hard to say exactly how it works in Unity since it is private. But they do make the disclaimer
`Note: This function is very slow. It is not recommended to use this function every frame. In most cases you can use the singleton pattern instead.` So I do imagine it could be something similar?

I did take a look at Unreal Engine as well since their code is open. I couldn't quite understand entirely how it was working. I think they are doing something much more clever to determine types but they make a similar disclaimer.
```
/** 
*	Find all Actors in the world of the specified class. 
*	This is a slow operation, use with caution e.g. do not use every frame.
*/
static ENGINE_API void GetAllActorsOfClass(const UObject* WorldContextObject, TSubclassOf<AActor> ActorClass, TArray<AActor*>& OutActors);
```
I have some ideas that I will look into to do this better based off of what I gathered from Unreal Engine, but that will take some time to flesh out.

pong.h
jonmgomez: Yep, makes sense, changed it back to unique_ptr

pong.h
cjappl: Yeah interesting. I think these likely both indicate a dynamic cast under the hood.

Definitely no rush on figuring any of it out, just a good thing to be aware of the trade offs :)

PR #8, opened 2023-11-09
Reogranize project into new subdirectories

jonmgomez:
Nice contained PR. Reorganizing the project into render, engine, and gameobject folders to better contain specific parts of the project. Include some new changes to cmake to build multiple libraries including an interface. How fancy.


CMakeLists.txt
```diff
@@ -7,50 +7,32 @@ set(CMAKE_CXX_STANDARD_REQUIRED 17)
...
-    shader.cpp
-    texture.cpp
-    vertexarray.cpp
-    vertexbuffer.cpp
-    vertexbufferlayout.cpp
-    wall.cpp
 )
 
-add_executable(pong ${SOURCES})
+add_executable(${PROJECT_NAME} ${SOURCES})
```
cjappl, line 12: Good idea!

CMakeLists.txt
```diff
@@ -7,50 +7,32 @@ set(CMAKE_CXX_STANDARD_REQUIRED 17)
...
+add_library(compile_flags INTERFACE)
 
 # I would use /Wall here but unfortunately the msvc headers trigger some pedantic warnings,
 # and I would like to keep /WX as well. Since I don't have an easy way to ignore the warnings
 # for the system libraries, just use /W4. Seems like /Wall has some weird extra warnings anyways.
-target_compile_options(pong PRIVATE /W4 /WX /MP)
+target_compile_options(compile_flags INTERFACE /W4 /WX /MP)
 
-target_compile_definitions(pong PRIVATE
+target_compile_definitions(compile_flags INTERFACE
```
cjappl, line 21: oh baby, now we are cmaking! This is a widely used technique.

engine/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+    pong.cpp
+    rectangle.cpp
+)
+
+add_library(engine ${SOURCES})
+
+target_link_libraries(engine
+    PUBLIC
+        render
+        gameobjects
```
cjappl, line 16: Do you understand the difference between public, private, and interface, and why each should be used?

cjappl (commented):
Looks great. Thanks for using `git mv` instead of normal mv to maintain sanity through a big one.

Use this opportunity to ensure you know the difference between the three specifiers PUBLIC, PRIVATE and INTERFACE, otherwise this looks great

engine/CMakeLists.txt
jonmgomez: Kind of, as far as I understood, public is used here to link these libraries, then anything that links engine would not have to link render or gameobjects because it would also link against them automatically, while private would not. And interface is a weird middle case which includes properties or libraries for a target in a public manner but does not use them for itself

cjappl: Yep. So I think the easiest way to understand these is in terms of headers vs source files.

If something is PRIVATE it means the classes and `#include` for that dependency are only in the source files (.cpp)

If something is interface, it means the classes and `#include` for that dependency are only in the header files (.h)

PUBLIC is both headers and source.


In practice, you want to ensure that as many dependencies are PRIVATE as possible. This should be the default you go with, and too many dependencies being PUBLIC is a code smell that you have some wacky dependency structure going on. This affects code architecture, as well as compile times. Technically, you could always just link everything as PUBLIC, but that would mean if you change a source file in module A, ALL of your other code needs to be recompiled. If everything is PRIVATE, changes are self contained and don't leak elsewhere, improving compile times and encapsulation.

cjappl: All public dependencies have the ability to cascade to other modules, so if you have foo PUBLIC depends on bar, and bar PUBLIC depends on baz, baz will link foo.

You can see how quickly that gets out of hand. All your modules know too much about each other and lead to architectural nastiness.

PR #9, opened 2023-11-12
Add timer callbacks for gameobjects functions

jonmgomez:
After the ball scores, things should wait for a minute before starting the game again. So, a timer is needed to call the ResetBall() function a few seconds after scoring. This Timer class holds requests involving a function callback and a timeout, keep track of when this should be called by keeping track of processing time between frames and call when ready.

Additionally, this introduces Timer::frameTime. This variable measures the processing time from the previous frame to this one. This is used in calulations to accomplish things over time to make them framerate independent. Ex. movement. A movement addition vector to position in OnUpdate() should be multiplied by this value otherwise running OnUpdate() more times (for higher framerates) will result in more additions to position in a second.

### DIFF
--- a/engine/pong.cpp
+++ b/engine/pong.cpp
@@ -91,9 +91,9 @@ void Pong::Cleanup()
     GetInstance().mGameObjects.clear();
 }
 
-void Pong::SetTimeout(int gameObjectId, int timeoutMs, std::function<void()> callback)
+void Pong::SetTimeout(int gameObjectId, std::chrono::duration<double> timeout, std::function<void()> callback)
 {
-    GetInstance().mTimer.AddTimer(gameObjectId, timeoutMs, callback);
+    GetInstance().mTimer.AddTimer(gameObjectId, timeout, callback);
 }
 
 } // namespace pong
--- a/engine/pong.cpp
+++ b/engine/pong.cpp
@@ -17,6 +17,16 @@ Pong& Pong::GetInstance()
     return instance;
 }
 
+CollisionManager& Pong::GetCollisionManager()
+{
+    return mCollisionManager;
+}
+
+Timer& Pong::GetTimer()
+{
+    return mTimer;
+}
+
 void Pong::Init()
 {
     auto player = std::make_unique<Player>();
@@ -62,8 +72,8 @@ void Pong::Init()
 
 void Pong::GameLoop()
 {
-    GetInstance().mTimer.HandleTimerCallbacks();
-    GetInstance().mCollisionManager.ProcessCollisions(GetInstance().mGameObjects);
+    Pong::GetInstance().GetTimer().HandleTimerCallbacks();
+    Pong::GetInstance().GetCollisionManager().ProcessCollisions(GetInstance().mGameObjects);
 
     for (auto& gameObject : GetInstance().mGameObjects)
     {
--- a/engine/pong.cpp
+++ b/engine/pong.cpp
@@ -62,8 +62,8 @@ void Pong::Init()
 
 void Pong::GameLoop()
 {
-    GetInstance().mCollisionManager.ProcessCollisions(GetInstance().mGameObjects);
     GetInstance().mTimer.HandleTimerCallbacks();
+    GetInstance().mCollisionManager.ProcessCollisions(GetInstance().mGameObjects);
 
     for (auto& gameObject : GetInstance().mGameObjects)
     {
--- a/engine/pong.cpp
+++ b/engine/pong.cpp
@@ -62,12 +62,13 @@ void Pong::Init()
 
 void Pong::GameLoop()
 {
+    GetInstance().mCollisionManager.ProcessCollisions(GetInstance().mGameObjects);
+    GetInstance().mTimer.HandleTimerCallbacks();
+
     for (auto& gameObject : GetInstance().mGameObjects)
     {
         gameObject->OnUpdate();
     }
-
-    GetInstance().mCollisionManager.ProcessCollisions(GetInstance().mGameObjects);
 
     for (auto& gameObject : GetInstance().mGameObjects)
     {
@@ -78,6 +79,11 @@ void Pong::GameLoop()
 void Pong::Cleanup()
 {
     GetInstance().mGameObjects.clear();
+}
+
+void Pong::SetTimeout(int gameObjectId, int timeoutMs, std::function<void()> callback)
+{
+    GetInstance().mTimer.AddTimer(gameObjectId, timeoutMs, callback);
 }
 
 } // namespace pong
--- a/engine/pong.h
+++ b/engine/pong.h
@@ -2,10 +2,12 @@
 
 #include "collisionmanager.h"
 #include "gameobject.h"
+#include "timer.h"
 #include "utils.h"
 
 #include <memory>
 #include <vector>
+#include <functional>
 
 namespace pong
 {
@@ -22,6 +24,7 @@ private:
 
     std::vector<std::unique_ptr<GameObject>>mGameObjects {};
     CollisionManager mCollisionManager {};
+    Timer mTimer {};
 
 public:
     static Pong& GetInstance();
@@ -43,6 +46,8 @@ public:
         }
         return nullptr;
     }
+
+    static void SetTimeout(int gameObjectId, int timeoutMs, std::function<void()> callback);
 };
 
 } // namespace pong
--- a/engine/pong.h
+++ b/engine/pong.h
@@ -5,6 +5,7 @@
 #include "timer.h"
 #include "utils.h"
 
+#include <chrono>
 #include <memory>
 #include <vector>
 #include <functional>
@@ -51,7 +52,7 @@ public:
         return nullptr;
     }
 
-    static void SetTimeout(int gameObjectId, int timeoutMs, std::function<void()> callback);
+    static void SetTimeout(int gameObjectId, std::chrono::duration<double> timeout, std::function<void()> callback);
 };
 
 } // namespace pong
--- a/engine/pong.h
+++ b/engine/pong.h
@@ -22,6 +22,9 @@ private:
     Pong& operator=(Pong&&) = delete;
     ~Pong() = default;
 
+    CollisionManager& GetCollisionManager();
+    Timer& GetTimer();
+
     std::vector<std::unique_ptr<GameObject>>mGameObjects {};
     CollisionManager mCollisionManager {};
     Timer mTimer {};
--- a/engine/pong.h
+++ b/engine/pong.h
@@ -25,6 +25,7 @@ private:
     std::vector<std::unique_ptr<GameObject>>mGameObjects {};
     CollisionManager mCollisionManager {};
     Timer mTimer {};
+    bool mFirstFrame { true };
 
 public:
     static Pong& GetInstance();
--- a/engine/timer.cpp
+++ b/engine/timer.cpp
@@ -1,5 +1,7 @@
 #include "timer.h"
 
+#include <spdlog/spdlog.h>
+
 #include <algorithm>
 #include <chrono>
 #include <iostream>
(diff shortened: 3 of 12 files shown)


engine/pong.cpp
```diff
@@ -62,13 +62,14 @@ void Pong::Init()
 
 void Pong::GameLoop()
 {
+    GetInstance().mTimer.HandleTimerCallbacks();
```
cjappl: I would possibly be encouraged to use the fully qualified version here, aka `Pong::GetInstance()` just for clarity.

I may also add a getter to `GetTimer` instead of interacting with the raw member variable (same with collision manager)

engine/pong.h
```diff
@@ -43,6 +47,8 @@ class Pong
         }
         return nullptr;
     }
+
+    static void SetTimeout(int gameObjectId, int timeoutMs, std::function<void()> callback);
```
cjappl: Nice use of std::function.

engine/timer.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include <algorithm>
+#include <chrono>
+#include <iostream>
+#include <memory>
+
+namespace pong
+{
+
+static constexpr int MS_TO_US = 1000;
+static constexpr int S_TO_US = 1000000;
```
cjappl: I would recommend using std::chrono instead of things like this. I think even the frameTime can be changed to some version of a chrono something

engine/timer.cpp
```diff
@@ -0,0 +1,50 @@
...
+        if (timerRequest->mTimeoutUs <= 0)
+        {
+            timerRequest->mCallback();
+        }
+    }
+
+    mActiveTimers.erase(std::remove_if(mActiveTimers.begin(), mActiveTimers.end(), [](const auto& timerRequest) {
+        return timerRequest->mTimeoutUs <= 0;
+    }), mActiveTimers.end());
+}
```
cjappl, line 34: Damn, look at all the modern cpp in this function. really a very different paradigm than spatial. I'm glad you're embracing it and learning it (regardless of if you like it or not). The experience with it is very valuable, and will be expected in many gigs.

engine/timer.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+    Timer::frameTime = static_cast<float>(timeWaited.count()) / static_cast<float>(S_TO_US);
+
+    const int timeWaitedUs = static_cast<int>(timeWaited.count());
+    for (auto& timerRequest : mActiveTimers)
+    {
+        timerRequest->mTimeoutUs -= timeWaitedUs;
+        if (timerRequest->mTimeoutUs <= 0)
+        {
+            timerRequest->mCallback();
```
cjappl: I recognize this timer pattern, seems familiar to an old codebase I used to work in

engine/timer.cpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+void Timer::AddTimer(int gameObjectId, int timeoutMs, std::function<void()> callback)
+{
+    // Interface uses ms but internally keeps track using us. This is just in case the
+    // game's framerate is uncapped. It is possible that {timeWaited} is less then 1ms
+    // and comes out to 0 then the timer will not proceed.
+    const int timeoutUs = timeoutMs * MS_TO_US;
+    auto timerRequest = std::make_shared<TimerRequest>(gameObjectId, timeoutUs, callback);
+    mActiveTimers.push_back(timerRequest);
```
cjappl: I don't know if you need to use a shared pointer here? You may just be able to push the TimerRequest object itself on there with no pointer wrapping it?

engine/timer.h
```diff
@@ -0,0 +1,39 @@
...
+#include <queue>
+#include <vector>
+
+namespace pong
+{
+
+struct TimerRequest
+{
+    int mGameObjectId { -1 };
+    int mTimeoutUs { -1 };
```
cjappl: would recommend at least a named constant. If this is something that may or may not be there, I would recommend possibly looking into `std::optional`, very convenient for avoiding weird sentinels like this

engine/timer.cpp
cjappl: If you do that, may be a good opportunity to use `std::move` here, or even `mActiveTimers.emplace_back`

Homework question: What does `emplace_back` do, and why would one consider it?

gameobjects/ball.cpp
```diff
@@ -3,16 +3,21 @@
...
+#include <spdlog/spdlog.h>
 
 #include <random>
 
 namespace pong
 {
 
 static constexpr float BALL_WIDTH = 20.0f;
-static constexpr float BALL_SPEED_BOUNCE_INCREMENT = 0.5f;
+static constexpr float BALL_SPEED_BOUNCE_INCREMENT = 75.0f;
```
cjappl, line 16: NOW WE'RE GAMING

gameobjects/ball.cpp
```diff
@@ -3,16 +3,21 @@
...
 static constexpr float Y_STARTING_POSITION_BOUNDS = 500.0f;
 
+static constexpr int BALL_RESET_WAIT_MS = 3000;
```
cjappl: could consider std::chrono

gameobjects/ball.cpp
```diff
@@ -28,15 +33,15 @@ void Ball::OnStart()
...
+    const glm::vec3 newPosition = GetPosition() + mVelocity * Timer::frameTime;
     SetPosition(newPosition);
 }
 
 void Ball::OnCollisionStart(GameObject& other)
 {
     if (other.GetInstanceName() == "ScoreArea")
     {
-        ResetBall();
+        SetTimeout(BALL_RESET_WAIT_MS, std::bind(&Ball::ResetBall, this));
```
cjappl: std::bind has fallen somewhat out of fashion. I think you should be able to get away with a lambda that captures `this`. Worth a try to see if it works, not sure the impact when using it with std::function

Aka:

```
SetTimeout(BALL_RESET_WAIT_MS, [this] ()
{
    this->ResetBall();
}
```

or something like that

cjappl (approved):

engine/timer.cpp
jonmgomez: I'm not sure what you mean by using std::chrono. Is it just a different type? Instead of `int MS_TO_US` it would be `std::chrono::some_type MS_TO_US`? 
And with frameTime, as long as it is easy to use with any calculation (like being multiplied to a vector), I could change that too. `SetPosition(GetPosition() + mVelocity * Timer::frameTime);`

engine/timer.cpp
jonmgomez: Yeah, so far it has been pretty fun to learn! Theres a lot of cool features in it. smart pointers are much, **much** nicer to use. Templates are sometimes annoying to deal with and figure out entirely, but they are very useful and I would much rather have them then not.

engine/timer.cpp
jonmgomez: Yeah that was a theory I had at one point, whether they need to be shared_ptr or not. I think I was just used to our old SplArray that forced everything to be allocated or it wouldn't work. Sometimes I just forget to test my theory 😁. I think I also had made this a shared_ptr because I was originally considering a separate thread for this timer class haha.

I looked at emplace_back briefly when I was setting up the objects in Pong::Init(), but I don't think I was able to because I needed to call OnStart() on the object after putting it in the array. Thats pretty cool to avoid the temporary object.
It seems like you would always try to use emplace_back though. Would there be any reason to not use emplace_back if you were able to?

gameobjects/ball.cpp
jonmgomez: Same comment as above, additionally would I have to change my function signature for SetTimeout to take a std::chrono type? Or at least overload it so it could?

gameobjects/ball.cpp
jonmgomez: It does work, very cool!

gameobjects/ball.cpp
jonmgomez: Sheeeesh!

engine/pong.h
jonmgomez: Yeah. It is really nice that this seemingly can callback any function with any amount of args using std::bind or a lambda? Makes something like this real easy to implement.

engine/timer.cpp
cjappl: Something like this:

```
    // Step 1: Define time in seconds
    std::chrono::seconds sec(5);

    // Step 2: Convert to milliseconds
    std::chrono::milliseconds millis = std::chrono::duration_cast<std::chrono::milliseconds>(sec);

    // Step 3: Output the milliseconds
    std::cout << "Milliseconds: " << millis.count() << std::endl;
```

By keeping all of your time in `chrono` you reduce the chance of messing up some conversion. Note there are no raw integers here, the library abstracts that away.

gameobjects/ball.cpp
cjappl: I think you would likely change the function signature to take the chrono type. The good thing is that you can then pass in ANY type of chrono (chrono Seconds for instance) and it would be automatically converted.

This way anyone could set a time in whatever units they wanted, and the interface would automagically make it work

engine/timer.cpp
```diff
@@ -25,26 +25,25 @@ void Timer::HandleTimerCallbacks()
...
 {
     // Interface uses ms but internally keeps track using us. This is just in case the
-    // game's framerate is uncapped. It is possible that {timeWaited} is less then 1ms
-    // and comes out to 0 then the timer will not proceed.
+    // game's framerate is uncapped. If using ms, it is possible that {timeWaited} is less then 1ms
+    // and comes out to 0. This will cause the timer to not proceed.
     const int timeoutUs = timeoutMs * MS_TO_US;
-    auto timerRequest = std::make_shared<TimerRequest>(gameObjectId, timeoutUs, callback);
-    mActiveTimers.push_back(timerRequest);
+    mActiveTimers.emplace_back(gameObjectId, timeoutUs, callback);
```
cjappl, line 42: nice!! much better. 

Any time you use "reference semantics" also known as "pointer semantics" there are a few things that can go wrong

1. performance is way worse, because you are fragmenting memory and can't keep it in the cache
2. You have to deal with the whole "is it valid" nullptr stuff, which can make your code gross
3. You have to explicitly determine ownership, which is it's own can of worms.


Wherever possible, prefer "value semantics" over reference semantics. It makes the code more comprehensible and easier to deal with overall.

cjappl (approved):

gameobjects/ball.cpp
jonmgomez: Ah! That is much nicer!

engine/timer.cpp
cjappl: emplace_back is generally the way to do things if you are building the object then immediately adding it to the collection! I don't think there is any reason not to use it in that case....

push_back is when you have something already constructed and you want to append

engine/pong.cpp
```diff
@@ -81,9 +91,9 @@ void Pong::Cleanup()
     GetInstance().mGameObjects.clear();
 }
 
-void Pong::SetTimeout(int gameObjectId, int timeoutMs, std::function<void()> callback)
+void Pong::SetTimeout(int gameObjectId, std::chrono::duration<double> timeout, std::function<void()> callback)
```
cjappl, line 94: You could take these by const reference i would guess. I don't think std::function it makes much of a difference, but the chrono object it is probably preferable

gameobjects/ball.cpp
```diff
@@ -16,7 +16,7 @@ static constexpr float BALL_WIDTH = 20.0f;
 static constexpr float BALL_SPEED_BOUNCE_INCREMENT = 75.0f;
 static constexpr float Y_STARTING_POSITION_BOUNDS = 500.0f;
 
-static constexpr int BALL_RESET_WAIT_MS = 3000;
+static constexpr std::chrono::seconds BALL_RESET_WAIT_S { 3 };
```
cjappl, line 19: nice

PR #10, opened 2023-11-13
Attempt to create a Real TIme Logger

jonmgomez:
This is my attempt making a real time logging thread. Going to put this PR as a draft for now. I think there is some funkiness here so I may want to redo parts of it.


engine/logger.h
```diff
@@ -0,0 +1,150 @@
...
+            break;
+        }
+    };
+
+private:
+    inline static int sLogNumber { 0 };
+};
+
+inline std::unique_ptr<Logger> gLogger { nullptr };
+inline std::unique_ptr<RealTimeLogger> gRTLogger { nullptr };
```
jonmgomez: I thought about making these classes a Singleton, but I feel a bit weird making too many classes Singletons. I think that is probably a code smell as well. 
This is done in a similar way as we did at Spatial. Is there a better way to handle these singleton-ish classes?

engine/realtimelogger.h
```diff
@@ -0,0 +1,49 @@
...
+    RealTimeLogger();
+    ~RealTimeLogger();
+
+    template<typename... Args>
+    void Log(int level, const char *format, Args... args)
+    {
+        LogData logData;
+        logData.mLevel = level;
+        logData.mSequenceNum = Logger::GetNextLogNumber();
+        fmt::format_to_n(logData.mMessage, MAX_LOG_MESSAGE_SIZE, format, args...);
```
jonmgomez: spdlog does not have an interface to get the final log message as a string or char*. So I have to use it eventually in the RT logger to get the timestamp and level formatted. My solution was to use fmt which spdlog uses anyway. Format the general message now and store that in the LogData. Then when we really need to log, the sequence number get prepended to the message and everything is forwarded to an spdlog call.

I did attempt to store the variadic arguments themselves to LogData so it can be passed to spdlog when ready. But that came out to be much uglier, and may not be possible entirely.

engine/realtimelogger.cpp
```diff
@@ -0,0 +1,66 @@
+// spdlog must be included before fmt. Since spdlog uses fmt, it will cause errors otherwise
+#include <spdlog/spdlog.h>
```
jonmgomez, line 2: Some funkiness because spdlog uses fmt under the hood. Not sure if there is a good way around it.

jonmgomez:
Added you @cjappl . No rush to take a look, it is just a draft and may need some changes. Let me know if there is anything non-real time safe I am doing here. 

Pretty cool to see it in action though. I tried benchmarking a little bit. 20 regular std::cout statements in the game loop drops the fps to 90. Pushing the real time logger to the limit, I tried 30 RT logs in the game loop and still kept a stable capped 144 fps!

engine/concurrentqueue.h
```diff
@@ -0,0 +1,3747 @@
+// Provides a C++11 implementation of a multi-producer, multi-consumer lock-free queue.
```
cjappl: I'd recommend adding this in tps with a similar pattern as you did before

engine/CMakeLists.txt
```diff
@@ -15,6 +16,7 @@ target_link_libraries(engine
     PUBLIC
         render
         gameobjects
+        fmt
```
cjappl: I would guess fmtlib has a `fmt::fmt` style.

When using the "fully qualified" name of a library, it helps you debug if something goes wrong, so they're generally recommended.

engine/logger.h
cjappl: Something like a logger it's probably fine enough to have these as global static singletons. Everyone needs to know about it in a sane way.

The very purist version would be to pass it in through every module, but that gets extremely messy, I don't think it's really advisable...

all companies I've been at have the logger be something like this

engine/logger.h
```diff
@@ -0,0 +1,150 @@
...
+void LogCritical(const char* format, Args... args)
+{
+    if (gLogger != nullptr)
+    {
+        gLogger->Log(spdlog::level::critical, format, args...);
+    }
+}
+
+template<typename... Args>
+void LogError(const char* format, Args... args)
```
cjappl, line 89: variadic template arguments, very spicy

engine/logger.h
```diff
@@ -0,0 +1,150 @@
...
+        gRTLogger->Log(spdlog::level::critical, format, args...);
+    }
+}
+
+template<typename... Args>
+void RealTimeLogError(const char* format, Args... args)
+{
+    if (gRTLogger != nullptr)
+    {
+        gRTLogger->Log(spdlog::level::err, format, args...);
```
cjappl: you may want to (for all these calls) do std::forward for the variadic args.

This is going to get really into the weeds of nasty template BS, but I think the purely correct version of something like this would be:

```


#include <utility> // for std::forward

template<typename... Args>
void RealTimeLogCritical(const char* format, Args&&... args)
{
    if (gRTLogger != nullptr)
    {
        gRTLogger->Log(spdlog::level::critical, format, std::forward<Args>(args)...);
    }
}

```

The very short explanation is this will preserve the "value category" of everything in the argument pack, that is rvalue references, lvalue references, const lvalue references, const rvalue references ........... will all remain the same way they are passed in, instead of being converted here.

There was a great talk at this year's cppcon on forwarding references that I recommend if you want to grok this.

I would not spend a ton of time wrapping your head around this unless you're really curious. I would recommend:

1. ignoring me, it doesn't really matter
2. trusting me, and just using it

either option totally fine, just wanted to bring it up

engine/logger.h
```diff
@@ -0,0 +1,150 @@
...
+        case spdlog::level::critical:
+            spdlog::critical(format, logNum, args...);
+            break;
+        default:
+            break;
+        }
+    };
+
+private:
+    inline static int sLogNumber { 0 };
```
cjappl: should be atomic

cjappl: also inline is probably overkill, unless you had a specific reason for it

engine/realtimelogger.h
```diff
@@ -0,0 +1,49 @@
...
+    int mSequenceNum { 0 };
+    char mMessage[MAX_LOG_MESSAGE_SIZE] {};
+};
+
+class RealTimeLogger
+{
+private:
+    static constexpr int MAX_LOGS = 100;
+
+    std::thread mThread {};
```
cjappl: consider using a scoped thread here and avoid having to have a destructor/constructor.

Alternatively, write or delete the move/copy constructors. If someone tries to copy/move right now you will get UB and a likely crash

engine/realtimelogger.h
```diff
@@ -0,0 +1,49 @@
...
+namespace pong
+{
+
+static constexpr int MAX_LOG_MESSAGE_SIZE = 512;
+
+struct LogData
+{
+    int mLevel { 0 };
+    int mSequenceNum { 0 };
+    char mMessage[MAX_LOG_MESSAGE_SIZE] {};
```
cjappl: std::array preferable

engine/realtimelogger.h
cjappl: Nice, this seems totally reasonable to me. This function seems realtime safe to my eye 👍

engine/realtimelogger.h
```diff
@@ -0,0 +1,49 @@
...
+    bool mAlive { true };
+
+    void Run();
+
+public:
+    RealTimeLogger();
+    ~RealTimeLogger();
+
+    template<typename... Args>
+    void Log(int level, const char *format, Args... args)
```
cjappl, line 30: you should consider making both of these (rtlogger::Log and normalLogger::Log) override some pure virtual interface. It will help them both match and changes will propagate to either of them

engine/realtimelogger.h
cjappl: after looking at the mAlive thing, I think you likely still need a destructor, (unless you want to get really crazy and write some kind of `ScopedTrueBool`

So what you could do here is:

1. Use a jthread
2. AND delete copy/move constructors


a reminder from rule of five "if you define one of them, you must define all of them". So by making this class have a destructor, you must write the rest of them.

Dave had an interview question literally two days ago from Apple where the answer was to complete the rule of five. Keep all this in mind :)

cjappl (approved):
Looks great, you must have learned from the best

cjappl:
> Added you @cjappl . No rush to take a look, it is just a draft and may need some changes. Let me know if there is anything non-real time safe I am doing here.
> 
> Pretty cool to see it in action though. I tried benchmarking a little bit. 20 regular std::cout statements in the game loop drops the fps to 90. Pushing the real time logger to the limit, I tried 30 RT logs in the game loop and still kept a stable capped 144 fps!

Love it, that means you did it right. Good on you for profiling it, makes a huge difference.

engine/realtimelogger.h
jonmgomez: I don't think I am able to make it a virtual function since it is a template https://stackoverflow.com/questions/2354210/can-a-class-member-function-template-be-virtual. Unless there is something I'm missing?

engine/realtimelogger.h
jonmgomez: So I kept the constructor and deleted the copy/move constructors. It seems that jthread is in C++20 though. I may want to move to C++20 at some point, but likely not now. I was very curious if I can use coroutines for callback instead of the timer class from earlier, so still open to trying eventually. Plus I tried to switch to C++20 and got some build errors so it will take some looking into lol.

engine/logger.h
jonmgomez: I'll trust ya, you've been right before once or twice 😁

engine/realtimelogger.h
cjappl: ah, cool!! Didn't realize it was the bleeding edge :) cppcon cooked my brain too many times, making me think I'm living in the future

engine/realtimelogger.h
cjappl: wow! Today I learned, thanks for the info, I had no idea :) 

At dolby and spatial (the codebases where I spent the most time) they were very template-phobic. Templates are by far the weakest part of my C++ repertoire, love learning more about them!

PR #11, opened 2023-11-18
Create new opponent ai, and classes to easily access config, random numbers, and difficulty settings

jonmgomez:
This introduces some slightly more complex opponent ai. The opponent will predict the balls position and move towards hit when the ball bounces. It also involves some error that is influcenced by the current difficulty settings.

This also includes some classes to more easily access data from the config file, and to more easily get random numbers.

### DIFF
--- a/engine/config.cpp
+++ b/engine/config.cpp
@@ -1,0 +1,53 @@
+#include "config.h"
+
+#include "renderutils.h"
+
+#include <spdlog/spdlog.h>
+
+#include <fstream>
+
+namespace pong
+{
+
+Config& Config::GetInstance()
+{
+    static Config instance;
+    return instance;
+}
+
+bool Config::LoadConfig(const std::string& filePath)
+{
+    std::ifstream jsonFile(filePath);
+    if (!jsonFile.is_open())
+    {
+        spdlog::error("Failed to open JSON config file: {}", filePath);
+        return false;
+    }
+
+    nlohmann::json jsonData;
+
+    try
+    {
+        jsonFile >> jsonData;
+    }
+    catch (const nlohmann::json::parse_error& e)
+    {
+        spdlog::error("JSON config parsing error: {}", e.what());
+        return false;
+    }
+
+    GetInstance().mConfig = jsonData;
+    return true;
+}
+
+std::optional<nlohmann::json> Config::GetJsonValue(const std::string& key)
+{
+    if (!GetInstance().mConfig.contains(key))
+    {
+        return {};
+    }
+
+    return GetInstance().mConfig[key];
+}
+
+} // namespace pong
--- a/engine/config.cpp
+++ b/engine/config.cpp
@@ -1,8 +1,6 @@
 #include "config.h"
 
-#include "renderutils.h"
-
-#include <spdlog/spdlog.h>
+#include "logger.h"
 
 #include <fstream>
 
@@ -20,7 +18,7 @@ bool Config::LoadConfig(const std::string& filePath)
     std::ifstream jsonFile(filePath);
     if (!jsonFile.is_open())
     {
-        spdlog::error("Failed to open JSON config file: {}", filePath);
+        LogError("Failed to open JSON config file: {}", filePath);
         return false;
     }
 
@@ -32,7 +30,7 @@ bool Config::LoadConfig(const std::string& filePath)
     }
     catch (const nlohmann::json::parse_error& e)
     {
-        spdlog::error("JSON config parsing error: {}", e.what());
+        LogError("JSON config parsing error: {}", e.what());
         return false;
     }
 
--- a/engine/config.h
+++ b/engine/config.h
@@ -1,0 +1,58 @@
+#pragma once
+
+#include <nlohmann/json.hpp>
+
+#include <optional>
+
+namespace pong
+{
+
+class Config
+{
+private:
+    nlohmann::json mConfig {};
+
+public:
+    static Config& GetInstance();
+    static bool LoadConfig(const std::string& filePath);
+
+    static std::optional<nlohmann::json> GetJsonValue(const std::string& key);
+
+    template<typename T>
+    static T GetValue(const std::string& key)
+    {
+        if (!GetInstance().mConfig.contains(key))
+        {
+            return T{};
+        }
+
+        try
+        {
+            return GetInstance().mConfig[key].get<T>();
+        }
+        catch(const std::exception&)
+        {
+            return T{};
+        }
+    }
+
+    template<typename T>
+    static T GetValue(const std::string& key, T defaultValue)
+    {
+        if (!GetInstance().mConfig.contains(key))
+        {
+            return defaultValue;
+        }
+
+        try
+        {
+            return GetInstance().mConfig[key].get<T>();
+        }
+        catch(const std::exception&)
+        {
+            return defaultValue;
+        }
+    }
+};
+
+} // namespace pong
--- a/engine/difficulty.cpp
+++ b/engine/difficulty.cpp
@@ -15,7 +15,7 @@ void Difficulty::SetLevel(Level level)
     sLevel = level;
 }
 
-std::string DifficultyString(Difficulty::Level level)
+std::string Difficulty::to_string(Difficulty::Level level)
 {
     switch (level)
     {
--- a/engine/difficulty.cpp
+++ b/engine/difficulty.cpp
@@ -1,0 +1,35 @@
+#include "difficulty.h"
+
+namespace pong
+{
+
+Difficulty::Level Difficulty::sLevel = Level::Normal;
+
+Difficulty::Level Difficulty::GetLevel()
+{
+    return sLevel;
+}
+
+void Difficulty::SetLevel(Level level)
+{
+    sLevel = level;
+}
+
+std::string DifficultyString(Difficulty::Level level)
+{
+    switch (level)
+    {
+        case Difficulty::Level::Easy:
+            return "Easy";
+        case Difficulty::Level::Normal:
+            return "Normal";
+        case Difficulty::Level::Hard:
+            return "Hard";
+        case Difficulty::Level::Insane:
+            return "Insane";
+        default:
+            return "Unknown";
+    }
+}
+
+} // namespace pong
--- a/engine/difficulty.h
+++ b/engine/difficulty.h
@@ -18,6 +18,7 @@ public:
 
     static Level GetLevel();
     static void SetLevel(Level level);
+    static std::string to_string(Difficulty::Level level);
 
 private:
     Difficulty() = delete;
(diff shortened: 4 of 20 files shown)


engine/config.cpp
```diff
@@ -0,0 +1,51 @@
...
+    {
+        LogError("JSON config parsing error: {}", e.what());
+        return false;
+    }
+
+    GetInstance().mConfig = jsonData;
+    return true;
+}
+
+std::optional<nlohmann::json> Config::GetJsonValue(const std::string& key)
```
cjappl, line 41: nice use of optional here. I am very infatuated with optional recently and it's ability to simplify interfaces

engine/config.h
```diff
@@ -0,0 +1,58 @@
...
+        if (!GetInstance().mConfig.contains(key))
+        {
+            return T{};
+        }
+
+        try
+        {
+            return GetInstance().mConfig[key].get<T>();
+        }
+        catch(const std::exception&)
```
cjappl, line 33: as a mental exercise, how would you do this without exceptions? Many codebases disallow them, so always good to know both ways

engine/config.h
```diff
@@ -0,0 +1,58 @@
...
+        {
+            return T{};
+        }
+    }
+
+    template<typename T>
+    static T GetValue(const std::string& key, T defaultValue)
```
cjappl, line 40: you're a regular templatizer now, look at you

engine/difficulty.cpp
```diff
@@ -0,0 +1,35 @@
...
+{
+    return sLevel;
+}
+
+void Difficulty::SetLevel(Level level)
+{
+    sLevel = level;
+}
+
+std::string DifficultyString(Difficulty::Level level)
```
cjappl: typically this method would be named "to_string", it's a special thing that other classes have as well, as an experienced C++ dev, it'd be one of the first things I'd try on a new class

cjappl: before you ask: yes C++ has messed up conflicts of style, snake_case vs CamelCase...

engine/difficulty.h
```diff
@@ -0,0 +1,35 @@
...
+    static Level GetLevel();
+    static void SetLevel(Level level);
+
+private:
+    Difficulty() = delete;
+    Difficulty(const Difficulty&) = delete;
+    Difficulty& operator=(const Difficulty&) = delete;
+    Difficulty(Difficulty&&) = delete;
+    Difficulty& operator=(Difficulty&&) = delete;
+    ~Difficulty() = delete;
```
cjappl, line 29: interesting that all of these are deleted. Can you explain why you did it this way? (not critiquing, just curious)

engine/difficulty.h
```diff
@@ -0,0 +1,35 @@
...
+
+    static Level sLevel;
+};
+
+std::string DifficultyString(Difficulty::Level level);
```
cjappl: I would expect this to possibly be a static method of the Difficulty class, or at least in a namespace Difficulty, such that you could do

```
pong::Difficulty::to_string(level)
```

engine/pong.cpp
```diff
@@ -91,6 +95,18 @@ void Pong::Cleanup()
     GetInstance().mGameObjects.clear();
 }
 
+GameObject* Pong::FindGameObjectByName(const std::string& name)
+{
+    for (auto& gameObject : GetInstance().mGameObjects)
+    {
+        if (gameObject->GetInstanceName() == name)
+        {
+            return gameObject.get();
```
cjappl: I think there is likely a std alg for this, find perhaps?

engine/random.cpp
```diff
@@ -0,0 +1,42 @@
+#include "random.h"
+
+namespace pong
+{
+
+static std::random_device rd;
+static std::mt19937 generator(rd());
```
cjappl: I would recommend adding this as a member of the class. Static can get messy with multithreading, I think it probably makes sense to have it be passed around in the obj itself

engine/random.cpp
```diff
@@ -0,0 +1,42 @@
...
+}
+
+template<>
+int Random::Range<int>(int min, int max)
+{
+    std::uniform_int_distribution<int> dist(min, max);
+    return dist(generator);
+}
+
+template<>
```
cjappl, line 31: can you explain to me (because I don't know) why the template braces are empty here? What is the reason for it?

engine/random.cpp
```diff
@@ -0,0 +1,42 @@
...
+static std::mt19937 generator(rd());
+
+template<typename T>
+T Random::Range(T min, T max)
+{
+    static_assert(false);
+}
+
+template<>
+int Random::Range<int>(int min, int max)
```
cjappl: naming nit: possibly 

```
randomGenerator.GetValueInRange()
```

a method called Range makes me think it will return the range

engine/timer.cpp
```diff
@@ -12,6 +12,12 @@ namespace pong
 
 float Timer::frameTime = 1.0f;
 
+void Timer::Init()
+{
+    const auto currentTime = std::chrono::system_clock::now();
+    Timer::frameTime = (std::chrono::duration_cast<std::chrono::duration<float>>(currentTime - mLastTime)).count();
```
cjappl, line 18: nice adjustment to chrono

you could save yourself some brain cells and fingers by doing a `using std::chrono` here, seeing as the std library just LOVES its nested namespace and insane templates

gameobjects/ball.cpp
```diff
@@ -85,12 +83,19 @@ void Ball::ResetBall()
 {
     mSpeed = BALL_START_SPEED;
 
-    const float yStartingPos = distYPos(generator);
+    const float yStartingPos = Random::Range(-Y_STARTING_POSITION_BOUNDS, Y_STARTING_POSITION_BOUNDS);
     SetPosition(glm::vec3(0.0f, yStartingPos, 0.0f));
 
-    const float xDir = distStartDir(generator) ? 1.0f : -1.0f;
-    const float yDir = distStartDir(generator) ? 1.0f : -1.0f;
+    const float xDir = Random::Bool() ? 1.0f : -1.0f;
```
cjappl, line 89: I think there is a way to do a random choice in the STL, you could investigate

In python, it would be something like:

```
random.choice([1, -1])
```

cjappl: alternatively, could be a good addition to your Random class, taking in any collection of any size and randomly selecting one out of it - fun engineering subproblem there in generics programming

gameobjects/opponent.cpp
```diff
@@ -1,52 +1,109 @@
...
+            break;
+
+        case Difficulty::Level::Hard:
+            decisionTimeLowerBoundMs = 0;
+            decisionTimeUpperBoundMs = 100;
+            baseMissChance = 0.05f;
+            baseRampUpChance = 0.2f;
+            break;
+
+        case Difficulty::Level::Insane:
```
cjappl, line 87: 2 ez, gg

gameobjects/player.cpp
```diff
@@ -39,7 +39,7 @@ void Player::OnUpdate()
 
 void Player::OnCollisionStart(GameObject& other)
 {
-    if (other.GetInstanceName() == "Wall")
+    if (other.GetInstanceName().find("Wall") != std::string::npos)
```
cjappl, line 42: nice

cjappl (approved):

engine/config.cpp
jonmgomez: Yeah, Im not using this function yet, but I had some ideas for some config values that may be different types that also could not exist. So I thought this would be the perfect use for it!

engine/config.h
jonmgomez: I suppose you could use specilization template functions to implement this function for each type. Then use the json's libraries own type checking ex  `mConfig[key].is_bool()` to see if it is good to return it. That would be my first approach at least.

engine/difficulty.h
jonmgomez: Yeah, it comes from my hesitancy to create another singleton class lol. I just made the level variable static so that it doesn't need to do the whole GetInstance() thing. At that point there isn't really a need in keeping the constructor around

engine/random.cpp
jonmgomez: Well I am not entirely certain myself, at least not past "it needs to be empty otherwise it wont compile." But I guess when using template specialization, since you are not using the generic T for the type name, there is no need to specify the type in the braces since you doing it later like here with `Range<float>` 
I do think it is helpful having it, since you learn at a glance that this is a template function, but it has just been specialized

engine/timer.cpp
jonmgomez: Yeah writing in that timer had some long lines to it haha. I just wasn't sure how recommended it is to use `using` in general. I know ideally you at least don't use `using std`

engine/random.cpp
cjappl: Seems reasonable to me!!


From Dr. GPT:


GPT

The feature you're referring to is called "template specialization" in C++. In the context of your code snippet, it's used to define a specific implementation of a template function for a particular data type, in this case, int.

Here's a brief explanation:

    Template Specialization: This is a C++ feature that allows you to define a specific implementation of a template class or function for a particular data type or a set of data types. It's useful when the general template implementation is not suitable or needs to be optimized for certain types.

In your example, template<> indicates that this is a specialization of the Random::ValueInRange template function for the type int. Because it's a specialization, you don't need to specify the template parameter (<int>) again in the function definition; it's implied by the context.

To learn more about this feature, you could research "C++ template specialization."

For your next turn, you might want to ask about specific use cases or examples of template specialization.

engine/random.h
```diff
@@ -8,18 +8,22 @@ namespace pong
...
-    Random() = delete;
+    Random();
     Random(const Random&) = delete;
     Random& operator=(const Random&) = delete;
     Random(Random&&) = delete;
     Random& operator=(Random&&) = delete;
-    ~Random() = delete;
+
+    std::random_device rd;
+    std::mt19937 generator;
```
cjappl, line 26: You possibly want to init these with empty braces? just to make sure they are initialized and not garbage 

```
std::random_device rd{}
```

engine/config.h
cjappl: Seems feasible

engine/timer.cpp
cjappl: There are some reasons to use it, one valid one is in the context of a single function when you have extreme use of some really long class name/namespace that you want to pair down and make less egregious, like this.

I don't love file wide uses of it, or uses that just shave off a couple characters. In general, you should write your code so it's easily READ not easily WRITTEN. You read code a billion times, and hopefully only write it once. A little over verbosity is good

cjappl (approved):

PR #12, opened 2023-11-23
Create font and text rendering. Show player and opponent scores visually.

jonmgomez:
Much more complicated than I expected, this introduces a way to render text to the screen. Using stb_truetype to load in a true type font file (.ttf) it does many calculations to get the offset each character needs to have from the top of the line, spacing from other characters, and etc. It creates a btimap image that is then saved to a texture. To actually render, each character of the string of text is represented by a simple Rectangle quad. Then the bitmap image is overlayed onto that quad as the texture. 

Now the score controller will dynamically create two text objects for each player's score. Upon scoring, the corresponding text object's text is changed, and the text object is essentially recalculated and recreated.

Possibly not the most efficient way to do this; by rereading the font file every time the text is changed, but this PR was getting a bit too large.


engine/pong.h
```diff
@@ -38,6 +39,10 @@ class Pong
 
     static GameObject* FindGameObjectByName(const std::string& name);
 
+    static Text* AddText(const std::string& text, const std::string& fontPath,
```
jonmgomez, line 42: This is for creating a text object on the fly instead of in PongInit(). These parameters are probably temporary until I figure out everything that Text needs to know when created. I think it also might be a good idea to do something like emplace_back for this with GameObjects or Text. That way the unique_ptr can be created for the vector, and a reference can still be passed back.

render/texture.h
```diff
@@ -1,43 +1,48 @@
...
+    static Texture CreateFromSolidColor(unsigned char r, unsigned char g, unsigned char b, unsigned char a);
+
     Texture() = default;
-    explicit Texture(const std::string& filePath);
-    ~Texture();
+    Texture(const unsigned char& imageData, int width, int height, GLenum format);
+    // Copy constructors are deleted because the renderer id is used to destroy the object in opengl
+    // in the destructor. The move operator gets around this by setting the renderer id to 0.
     Texture(const Texture&) = delete;
-    Texture(Texture&&) = delete;
```
jonmgomez, line 15: So this is kind of interesting. Initally I had a lot of the copy/move constructors deleted for these OpenGL render classes. But I wanted to change the Texture for a Mesh to a stack variable instead, so I had to implement them all. However that caused a problem specfically when copying these objects.

Because they have a mRendererId that is created when the object on the GPU side is created. 
`glGenBuffers(1, &mRendererID);`
The destructor frees that object on the GPU side as well using that id
`glDeleteBuffers(1, &mRendererID);`

If the object is copied, the object it was copied from may destruct, and free the object from under us. So the only way for it to work without serious problems is to only implement the move constructors, and leave the copy constructors deleted. Because the move constructor will try to delete this object with an id of 0 which won't do anything
```
mRendererID = other.mRendererID;
other.mRendererID = 0;
```

render/vertexarray.h
```diff
@@ -11,11 +11,11 @@ class VertexArray
 {
 public:
     VertexArray();
-    ~VertexArray();
     VertexArray(const VertexArray&) = delete;
-    VertexArray(VertexArray&&) = delete;
     VertexArray& operator=(const VertexArray&) = delete;
-    VertexArray& operator=(VertexArray&&) = delete;
+    VertexArray(VertexArray&&);
+    VertexArray& operator=(VertexArray&&);
```
jonmgomez, line 17: So this is kind of interesting. Initally I had a lot of the copy/move constructors deleted for these OpenGL render classes. But I wanted to change the Texture for a Mesh to a stack variable instead, so I had to implement them all. However that caused a problem specfically when copying these objects.

Because they have a mRendererId that is created when the object on the GPU side is created. 
`glGenBuffers(1, &mRendererID);`
The destructor frees that object on the GPU side as well using that id
`glDeleteBuffers(1, &mRendererID);`

If the object is copied, the object it was copied from may destruct, and free the object from under us. So the only way for it to work without serious problems is to only implement the move constructors, and leave the copy constructors deleted. Because the move constructor will try to delete this object with an id of 0 which won't do anything
```
mRendererID = other.mRendererID;
other.mRendererID = 0;
```

render/texture.h
```diff
@@ -1,43 +1,48 @@
...
 {
 
 class Texture
 {
 public:
+
+    static Texture CreateFromFile(const std::string& filePath);
+    /* Assumes the data is grayscale and stored from top left to bottom right in vector */
+    static Texture CreateFromFontCharacter(const std::vector<unsigned char>& imageData, int width, int height);
+    static Texture CreateFromSolidColor(unsigned char r, unsigned char g, unsigned char b, unsigned char a);
```
jonmgomez: Curious if you have any thoughts (positives, negatives) on this kind of API. I didn't want to keep inheriting from this class like I did for SolidColorTexture. ie some kind of `FontTexture` for create from font bitmaps specfically. So instead I thought these static creation functions could avoid that. 

The only thing that is a little weird is how you cant really create smart pointers of a Texture easily, unless I were to create new static functions that would specifcally return that?

render/texture.h
```diff
@@ -1,43 +1,48 @@
...
+    ~Texture();
 
     void Bind(unsigned int slot = 0) const;
     void Unbind() const;
 
     int GetWidth() const;
     int GetHeight() const;
 
+    static std::vector<unsigned char> ConvertAlphaImageToRGBA(const std::vector<unsigned char>& imageData);
+    static void FlipImageVertically(const std::vector<unsigned char>& imageData, int width, int height, int comp, std::vector<unsigned char>& outFlippedImageData);
```
jonmgomez: This was me trying a couple different ways to hand back a large vector of data. The first returns a stack allocated std::vector by creating a new one in the function, fillling it with data, and std::move()'ing as the return value. I imagine that is better than straight returning by value, but I also have to imagine for large vectors (which images can be sometimes) it could still get pretty expensive?

The alternative I tried was passing in a stack variables and changing it in the function as an out variable. I find that kind of interface a little weird too though. I guess another option would be to pass back a smart point of a newly created std::vector? 
What do you think about these interfaces?

tps/CMakeLists.txt
```diff
@@ -38,3 +38,12 @@ FetchContent_Declare(concurrentqueue
...
 FetchContent_MakeAvailable(concurrentqueue)
+
+FetchContent_Declare(stb
+    URL ${CMAKE_SOURCE_DIR}/tps/stb-master.zip
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND ""
+)
+FetchContent_MakeAvailable(stb)
+add_library(stb INTERFACE)
+target_include_directories(stb INTERFACE ${stb_SOURCE_DIR})
```
jonmgomez, line 49: This was some code I found to compile the stb library manually since it is header only and does not have a CMakeLists.txt to it. Is there anything you woud do differently here?

engine/text.cpp
```diff
@@ -0,0 +1,180 @@
...
+Text::Text(const std::string& text, const std::string& path, float scale, int pixelLineHeight) :
+    mText { text },
+    mFontPath { path },
+    mScale { scale },
+    mPixelLineHeight { pixelLineHeight }
+{
+    CreateText();
+}
+
+void Text::CreateText()
```
jonmgomez, line 22: After writing this function. I must say, text rendering is probably my least favorite part of graphics rendering pipelines 🤣

engine/pong.cpp
```diff
@@ -88,11 +90,17 @@ void Pong::GameLoop()
     {
         gameObject->Render();
     }
+
+    for (auto& text : GetInstance().mTexts)
+    {
+        text->Render();
```
cjappl, line 96: The game text is stored in the global instance? That may be a bit of encapsulation breaking.

You may want to consider putting these in their own object that is explicitly passed around, I'll see how it looks through the rest of the review

engine/text.cpp
```diff
@@ -0,0 +1,180 @@
...
+{
+    std::ifstream file(mFontPath, std::ios::binary);
+    if (!file)
+    {
+        RealTimeLogError("Failed to open true type font file: {}", mFontPath);
```
cjappl, line 27: Is this a critical error that is unrecoverable? If so I would recommend just asserting/dying

render/texture.h
cjappl: Hmm, great question.

I think there are many ways to go about this process, so know this isn't the only way you could do it.

I would start by making an rgba struct, seeing as that's a pretty common thing here, and you can simplify the rest of your code with it.

From there, I would do the following:

* Drop the static methods, just use constructors, aka `Texture(const std::vector<unsigned char>& imageData, int width, int height)` and `Texture(const Rgba& color);`

from there, you can either call the constructor directly, or now make_unique or make_shared will defer to these constructors depending on their arguments.


I think despite not having a "nicely named" static function, the ability to naturally use make_shared or make_unique is a good simplification which doesn't have you exploding the complexity of:

static Texture CreateFromFontCharacter(const std::vector<unsigned char>& imageData, int width, int height);

// put this in a shared or unique pointer!!
static Texture* AllocateFromFontCharacter(const std::vector<unsigned char>& imageData, int width, int height);

and so on.


I think using the existing facilities may be an improvement

render/texture.h
cjappl: Alright, a couple things:

> and std::move()'ing as the return value

This should never be done. The things to investigate are "return value opitimization" and "copy elision". Basically, your friend the compiler does this automatically and you can mess it up by `move`ing a return value explicitly.

> for large vectors (which images can be sometimes) it could still get pretty expensive?

Moving a vector is actually EXTREMELY cheap. :) What is a vector, under the hood? What was splArray under the hood? Really, just a pointer to some data and a size.

If you are ready to look into the absolute insanity of an STL implementation, check out clang's version [here](https://github.com/llvm-mirror/libcxx/blob/78d6a7767ed57b50122a161b91f59f19c9bd0d19/include/vector#L342-L344). If you aren't busy being blinded by underscores and instanity, you'll notice there are only 3 members here, two pointers to beginning and end, and a pointer to an allocator.

Even if your vector has 10M elements, moving those elements is 3 pointer swaps. That is why std::move is SO POWERFUL. Really insane. Instead of 10M copies, you can do 3 swaps of the datatype your machine is most optimized to swap. Crazy.


> pass back a smart point of a newly created std::vector

So one thing to realize about anything heap allocated is that you now have to concern yourself with speed of the cache. In general, stack allocated variables are WAY more performant than heap allocated variables, and this has to do with the cache. (let me know if you want some sources/educational sources on this, good things to know). Basically, anything you can put on the stack, you should. Any benefit that you get by returning a pointer in terms of that specific copy is going to be shredded by your cache getting ripped up. We also just saw that moving a big vector is basically free and is O(1). 

I would recommend returning this value on the stack, then just being mindful to pass by reference and move instead of copy !

static std::vector<unsigned char> outFlippedImageData FlipImageVertically(const std::vector<unsigned char>& imageData, int width, int height, int comp );

render/vertexarray.h
cjappl: Good catch, and reasonable solution!

I guess the way to make the compy constructor would be to create a new object with a new ID for the new object, and not ever share the old ID

tps/CMakeLists.txt
cjappl: Seems reasonable to me. You can look at my rt_log library to see how I did it, I suspect I did pretty much the same thing

cjappl (approved):

engine/pong.cpp
jonmgomez: Not sure what you mean entirely. I think of the text objects like game objects. Something with a position and a mesh to render each frame. But they don't need the behavior from OnStart() or OnUpdate(). In order to be rendered every frame like the gameobjects do the global instance needs to render them every game loop through its list of Text objects

render/texture.h
jonmgomez: Yeah that is a good point.

render/texture.h
jonmgomez: Ah very interesting I figured everything gets implicitly copied when just returned like that. What a cool optimization.

engine/pong.cpp
cjappl: Yeah that's fine. I just always worry about adding things to global singleton objects. I think as a "after your done" task, you could dig in to unreal or unity and see how they collect renderables. Could be illuminating

PR #13, opened 2023-11-28
Implement sounds and audio mixing

jonmgomez:
Here is my naive approach to adding sound output to the game. I am using portaudio for my audio callback. Sounds can be held on gameobjects then played through the `PlaySound()` function. This add a new playing sound to the audio mixer. In each callback it reads from the audio file and mixes any together. Playing sounds just hold an index for the current sample it is reading, and a pointer to the Sound which holds the audio file data. 

I did also add some code for spatializing the audio based on where the audio play position is. This is enabled via config, because I just wanted to experiment with it a bit.


engine/audiomixer.cpp
```diff
@@ -0,0 +1,167 @@
...
+{
+    mPlayingSounds.emplace_back(sound);
+}
+
+void AudioMixer::PlaySound(const Sound& sound, const glm::vec3& position)
+{
+    mPlayingSounds.emplace_back(sound, position);
+}
+
+int AudioMixer::AudioCallback(const void* /*inputBuffer*/, void* outputBuffer,
```
jonmgomez, line 104: So a couple questions about the audio callback loop. I measured the time between callbacks and it follows a strange pattern. I would expect it to be around every 20ms right? But looking at it the time between callbacks is always something like

`~20000us, ~300us, ~300us, 20000us, 300us, 300us, and so on...`

Here is a stackoverflow question on the same thing. It gives some insight onto why, but I wasn't able to get a definitive answer. https://stackoverflow.com/questions/23214614/time-between-callback-calls

I guess I am mostly curious if you have seen something like this before? And more so, how does this effect question 2?

jonmgomez: And if I remeber correctly at spatial, we had the audio processing done on a timer using the event queue. So the scene enqueue a render scene timer for 20ms from now and then it would mix all the sounds in the scene. I guess I am confused as to how that was working with the pa callback? Was the buffer just assumed to be ready by the scene render timer, and then simply just put into the pa callback buffer? Does it make sense to do something like that here? 

 And if I do use something like this, does the audio processing time not only have to take <20ms every time, but also <300us every 2nd and 3rd time? Let me know if it is easier to explain this over zoom or something that might be easier.

engine/audiomixer.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+void AudioMixer::PlaySound(const Sound& sound)
+{
+    mPlayingSounds.emplace_back(sound);
+}
+
+void AudioMixer::PlaySound(const Sound& sound, const glm::vec3& position)
+{
+    mPlayingSounds.emplace_back(sound, position);
```
jonmgomez, line 100: I'll fix this when next I change things for this PR. I don't think this is quite thread safe right now. I tried to look into using a thread sanitizer, but it seems msvc has no support for one at all 😞

engine/audiomixer.cpp
```diff
@@ -0,0 +1,167 @@
...
+    LogInfo("Spatial Audio: {}", mSpatialAudioEnabled ? "Enabled" : "Disabled");
+
+    PaError err = Pa_Initialize();
+    if (err != paNoError) {
+        LogError("PortAudio initialization failed: {}", Pa_GetErrorText(err));
+    }
+
+    PaStreamParameters outputParameters;
+    outputParameters.device = Pa_GetDefaultOutputDevice();
+    outputParameters.channelCount = 2; // Stereo
```
cjappl: Better than a comment would be a named constant: kAudioChannelsStereo

engine/audiomixer.cpp
```diff
@@ -0,0 +1,167 @@
...
+    outputParameters.sampleFormat = paFloat32;
+    outputParameters.suggestedLatency = Pa_GetDeviceInfo(outputParameters.device)->defaultLowOutputLatency;
+    outputParameters.hostApiSpecificStreamInfo = nullptr;
+
+    err = Pa_OpenStream(&mStream, nullptr, &outputParameters, SAMPLE_RATE, FRAMES_PER_BUFFER, paClipOff, AudioCallbackWrapper, nullptr);
+    if (err != paNoError) {
+        LogError("PortAudio stream opening failed: {}", Pa_GetErrorText(err));
+        Pa_Terminate();
+        return false;
```
cjappl: could be an opportunity to just assert and die here. Seems like not opening the device is a critical failure and everything should be aborted immediately, same with below

engine/audiomixer.cpp
cjappl: embarrassing for msvc. Very unfortunate.

You may be able to compile with llvm on windows?? Not sure...

If mPlayingSounds is being used here and in the audio thread, it is true that this is not thread safe. The naive approach that you could start with is just adding a lock, but as I've long ranted about that isn't realtime safe ;)

engine/audiomixer.cpp
cjappl: Question 1:

That seems reasonable enough. As an audio callback user, all you concern yourself with is providing the right number of samples back. Everything internal to PA you have to assume they are doing their best to give you the best performance and latency they possibly can. In this case, it may mean wacky "inconsistent" callbacks. Who knows why, but you have to trust the abstraction.

You may be able to mess around with this by tweaking this value, but not sure:

```
outputParameters.suggestedLatency = Pa_GetDeviceInfo(outputParameters.device)->defaultLowOutputLatency;
```

Question 2:
Your callback should take bascially no time at all, if you've done it correctly. If you get a very low time, and it is realtime safe, and you're still having issues, you should increase your latency as described above.

If you want a history of what was going on between the spatializer and the PA callback at spatial, more than happy to talk it over on zoom. It's a little complex to type (and ugly)

engine/audiomixer.cpp
```diff
@@ -0,0 +1,167 @@
...
+{
+    float* out = static_cast<float*>(outputBuffer);
+
+    int frames = static_cast<int>(framesPerBuffer);
+    for (int i = 0; i < frames; i++)
+    {
+        float sampleL = 0.0f;
+        float sampleR = 0.0f;
+
+        for (auto& sound : mPlayingSounds)
```
cjappl, line 120: Definitely a data race with these sounds at the moment. This would be a good question for Dave on how to handle this properly.  This is how easy it is to get into very nasty situations with threading and realtime safety, as they often conflict.

My initial thought on this is to use a a try_lock, but this is going to get quite complicated quickly. To not bog you down in this detail, I would recommend possibly just using a plain old mutex for now, and file an issue to deal with this later. 

This is how easy it is to fall of the "realtime safe" wagon :) 


The solution of how to do this properly likely exists in this talk:
https://www.youtube.com/watch?v=Q0vrQFyAdWI

engine/audiomixer.cpp
```diff
@@ -0,0 +1,167 @@
...
+        sampleL = std::clamp(sampleL, -1.0f, 1.0f);
+        sampleR = std::clamp(sampleR, -1.0f, 1.0f);
+
+        *out++ = sampleL;
+        *out++ = sampleR;
+    }
+
+    mPlayingSounds.erase(std::remove_if(mPlayingSounds.begin(), mPlayingSounds.end(), [](const auto& sound) {
+        return sound.IsFinished();
+    }), mPlayingSounds.end());
```
cjappl, line 170: Not realtime safe, unfortunately :( (also not thread safe)

again, I would consider skipping this for now as the fix is going to be complex and painful :) yay audio

engine/playingsound.cpp
```diff
@@ -0,0 +1,50 @@
+#include "playingsound.h"
+
+namespace pong
+{
+
+PlayingSound::PlayingSound(const Sound& sound) :
+    mSound(&sound),
```
cjappl, line 7: can mSound be just saved as a reference?

engine/playingsound.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+glm::vec3 PlayingSound::GetPosition() const
+{
+    return mPosition;
+}
+
+bool PlayingSound::IsPositional() const
+{
+    return mPositional;
+}
```
cjappl, line 48: Nice abstraction for a sound! Good class design

engine/pong.cpp
```diff
@@ -69,6 +74,7 @@ void Pong::Init()
     GetInstance().mGameObjects.push_back(std::move(scoreController));
 
     Pong::GetInstance().GetTimer().Init();
+    ASSERT(Pong::GetInstance().GetAudioMixer().Init());
```
cjappl: I'd possibly be inclined to put the asserts down in this function and die immediately at the source of the problem, but this is OK as well!

cjappl (approved):

engine/audiomixer.cpp
jonmgomez: Yeah Ill just put a lock around these for now.

engine/audiomixer.cpp
jonmgomez: my fav 😁

engine/playingsound.cpp
jonmgomez: I figured pointers are preferable to references for member variables because they can be initalized to null? Rather than some garbage default value for a reference?

engine/pong.cpp
jonmgomez: Yeah, that does bring more immediate attention to where it failed

engine/audiomixer.cpp
jonmgomez: Okay cool. In that case I will probably just stick with what I got.

engine/playingsound.cpp
cjappl: ah, yes. And famously std::optional doesn't support references I don't think.

Fine to use pointers when you need a null :) 

Although is it ever valid for a playing sound to have a null sound? that seems like it would be malformed

engine/audiomixer.cpp
cjappl: Ok, talked to dave today and brought this up. There is a likely simple way to get these to be realtime safe.

Basically, you need a realtime safe queue (which you already have) but between your main thread and your audio thread. This queue would be full of sound event objects. Things like "StartPaddleSound", "StopPaddleSound" etc

In the realm of your audio thread, you have a static array of all the sounds you wish to play, and on a StartPaddleSound, you simply play the paddle sound file. 

The only thing that crosses the thread boundaries is the event, and that is thread safe in your lock free queue. The sound thread deals with interpreting the event, then choosing and playing the right sound until it hears the "stop" event.


You do not have to create this now, just writing this here for posterity if you ever feel a need to circle back to implement this. This is more akin to what normal game engines do, I would think :)

engine/playingsound.cpp
jonmgomez: I guess not, I was more worried about there being some garbage default value in the reference in case it did get used before it was initalized.

engine/audiomixer.cpp
jonmgomez: Cool, I'll keep that in mind for later 🧠

PR #14, opened 2023-12-01
Add in a title screen, and ui button events

jonmgomez:
A few things here:
- UIElements are a general base class that I envision can have different events based on what ui it is (button, sliders, checkboxes, 
etc)
- This adds in a title screen before the main game. It is layed out with some text, a button to start the game, go to settings (eventually), and exit the game. This is done by attaching callback listeners to ui events. When the ui manager register an event on a ui element it will callback all listeners of that event.
- Ability to load in different scenes by a call to `Pong::LoadNextScene()`. The caveat here is that it has to finish a full gameloop before swapping things, as trying to destruct and cleanup all objects while iterating through things causes many issues.

This is a bit large, but there is also a lot of boilerplater. And to build scenes and ui elements with code takes quite a bit (I miss doing this graphically).


render/renderutils.h
```diff
@@ -16,6 +16,36 @@ namespace pong
...
+struct GLRGBAColor
+{
+    float r { 1.0f };
+    float g { 1.0f };
+    float b { 1.0f };
+    float a { 1.0f };
+
+    GLRGBAColor() = default;
+
+    constexpr GLRGBAColor(float r, float g, float b, float a)
```
jonmgomez: Is there any reason you wouldn't mark these struct constructors with constexpr? In this case I only did because I was using it. constexpr is still a bit unknown to me, I don't see it too often.

engine/mesh.h
```diff
@@ -17,11 +18,14 @@ class Mesh
...
     virtual void Draw(const glm::vec3& position) const;
 
+    void SetColor(const GLRGBAColor& color);
+
 protected:
     VertexArray mVA {};
     VertexBuffer mVB {};
     IndexBuffer mIB {};
     Texture mTexture {};
+    GLRGBAColor mColor { 1.0f, 1.0f, 1.0f, 1.0f };
```
cjappl: I would possibly encourage a constant for white or similar, seems like something that will get reuse

engine/pong.h
```diff
@@ -40,15 +53,27 @@ class Pong
 
     static GameObject* FindGameObjectByName(const std::string& name);
 
-    static Text* AddText(const std::string& text, const std::string& fontPath,
-                         float scale, int pixelLineHeight = 128,
-                         const glm::vec3& position = glm::vec3(0.0f));
+    template<typename UIType, typename... Args>
+    static UIType* AddUIElement(Args&&... args)
+    {
+        static_assert(std::is_base_of<UIElement, UIType>::value, "UIType must derive from a UIElement");
```
cjappl: Ok, perfect opportunity for something fun here. 

The pattern you should use is called "SFINAE" which translates to "Substitution Failure is Not An Error". 

Without going into too much detail, in C++ before 20, there is a way to set bounds to types in a way that is like this, but provides much nicer compiler errors naturally, you may as well write one so you get use to it :) 

In fact, CppCoreGuidelines says you should ALWAYS specify bounds for every argument in a template using SFINAE in pre C++20, or concepts afterwards.


The way you would do this for this function using SFINAE is:

```
template<typename UIType, typename = std::enable_if_t<std::is_base_of_v<UIElement, UIType>>, typename... Args>
static UIType* AddUIElement(Args&&... args)
{
    // Function implementation
}
```

This will prevent any compilation if there it does NOT match the clause in the `enable_if`, and present you with a helpful diagnostic.

In C++20, we have concepts, which is the slightly (??) more sane way of doing it, it would look something like this:

```
template<typename T>
concept IsDerivedFromUIElement = std::is_base_of<UIElement, T>::value;

template<IsDerivedFromUIElement UIType, typename... Args>
static UIType* AddUIElement(Args&&... args)
{
    // Function implementation
}
```

Same exact result, you restrict the template type to conforming to some bounds.

engine/rectangle.cpp
```diff
@@ -23,6 +23,15 @@ static const std::array<unsigned int, 6> kIndicies = {
 Rectangle::Rectangle(float width, float height) :
     mWidth {width},
     mHeight {height}
+{
+    RecomputeVertices();
+
+    const unsigned char kColorValue = 255;
+    const RGBAColor allWhiteColor{kColorValue, kColorValue, kColorValue, kColorValue};
```
cjappl: As above, just abstract this out and you can use it anywhere /everywhere

engine/uieventmanager.cpp
```diff
@@ -0,0 +1,50 @@
...
+            if (Input::IsMouseButtonPressed(GLFW_MOUSE_BUTTON_LEFT))
+            {
+                if (!button.IsPressed())
+                {
+                    button.OnEvent(ButtonEvent::Pressed);
+                }
+            }
+            else if (button.IsPressed())
+            {
+                button.OnEvent(ButtonEvent::Release);
```
cjappl, line 40: Is this correct, on button pressed send event released?

engine/uieventmanager.h
```diff
@@ -0,0 +1,17 @@
...
+#include <memory>
+#include <vector>
+
+namespace pong
+{
+
+class UIEventManager
+{
+public:
+    void ProcessEvents(const std::vector<std::unique_ptr<UIElement>>& uiElements);
```
cjappl: Not sure if you use this type a lot, but could consider aliasing it

```
using UiElementCollection = std::vector<.....>
```

May save you some eye damage

render/renderutils.h
cjappl: I think this is the perfect class to do some constexpring with.

Basically what you need to know is constexpr is "this COULD BE evaluated at compile time, if the inputs are known at compile time. In that case, you incur a small compile time cost at the benefit of runtime performance.

This trade off depends on the context, but I think simple dataclasses like this are a general gimme for doing it

ui/button.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+UIElementType Button::GetType() const
+{
+    return UIElementType::Button;
+}
+
+void Button::AddListener(ButtonEvent event, std::function<void()> callback)
+{
+    ASSERT(event < ButtonEvent::EVENTS_COUNT);
```
cjappl, line 27: is this even possible? As long as someone isn't getting crazy and doing the cast to and from integers, this would be hard. `enum class`, which you've been good about using is much more resilient to issues like this than c style enums.

ui/button.cpp
```diff
@@ -0,0 +1,83 @@
...
+    mListeners.at(static_cast<int>(event)).emplace_back(callback);
```
cjappl, line 28: nice use of .at, love bounds checking vs normal []

ui/uielement.cpp
```diff
@@ -0,0 +1,33 @@
+#include "uielement.h"
+
+#include "pong.h"
+
+namespace pong
+{
+
+void UIElement::Render() const
+{
+}
```
cjappl: Empty methods are normally a little bit of a code smell. It may be better here to force implementation with a pure virtual than accidentally not implement this for a subclass

cjappl (approved):

ui/uielement.cpp
jonmgomez: Yeah that makes sense. Im not entirely sure of all the UIElements I need yet, but they probably need to implement render anyway

ui/button.cpp
jonmgomez: Well I guess this would just be preventing from doing something like AddListener(ButtonEvent::EVENTS_COUNT, []() {});
directly passed into the function.

engine/uieventmanager.cpp
jonmgomez: Yes, although that is confusing. This is similar to that collision detection system I have. If they are just now colliding call Enter, if they already were call Stay(), if they were but ar no longer call Stop(), I renamed this to WasPressed so it more clear that it was being pressed and no longer is, so release should be called.

engine/pong.h
jonmgomez: Aha that is awesome. Not only is it a better compiler error, but my editor can actually detect it before being compiled!

PR #15, opened 2023-12-06
Add settings screen, and slider and checkbox ui elements.

jonmgomez:
Pretty similar to the last PR, the new ui elements of the slider and checkbox allow to get some user input in a nice form. This lays out the settings page to tweak volume settings, difficulty and likely some more settings such as resolution in the future.

A couple other changes to note:
- Input class is changed to be able to distinguish between inputs being held, and first being pressed. This is so buttons and ui elements only trigger when hovering over them and the frame they were pressed.
-- Pressed - Frame the input was pushed down,
-- Held - Any frames after pressed when the input is still down
-- Released - Frame the input was released
-- NotPressed - Any frames after released when the input is still up
- Settings screen and title screen inherit from a UIMenu class. This is just for setting up similar animations for the buttons and text combos on both screens
- Both slider and checkbox use a MeshComponent to build a simple border around the ui element. This consists of 4 Rectangle meshes to layout the border.

### DIFF
--- a/engine/input.cpp
+++ b/engine/input.cpp
@@ -35,23 +35,23 @@ void Input::Init(GLFWwindow* window)
 
 void Input::Update()
 {
-    const auto updateInputState = [](InputState* inputStates, size_t size)
+    const auto updateInputState = [](auto& inputStatesArray)
     {
-        for (int i = 0; i < size; ++i)
+        for (auto& state : inputStatesArray)
         {
-            if (inputStates[i] == InputState::Pressed)
+            if (state == InputState::Pressed)
             {
-                inputStates[i] = InputState::Held;
+                state = InputState::Held;
             }
-            else if (inputStates[i] == InputState::Released)
+            else if (state == InputState::Released)
             {
-                inputStates[i] = InputState::NotPressed;
+                state = InputState::NotPressed;
             }
         }
     };
 
-    updateInputState(mKeys.data(), mKeys.size());
-    updateInputState(mMouseButtons.data(), mMouseButtons.size());
+    updateInputState(mKeys);
+    updateInputState(mMouseButtons);
 }
 
 void Input::KeyCallback(GLFWwindow* /*window*/, int key, int /*scancode*/, int action, int /*mods*/)
--- a/engine/input.cpp
+++ b/engine/input.cpp
@@ -11,14 +11,47 @@ static constexpr int WINDOW_WIDTH = 1280;
 static constexpr int WINDOW_HEIGHT = 960;
 
 // All keys are intalized to false, for not pressed
-std::array<bool, GLFW_KEY_LAST + 1> Input::mKeys {};
-std::array<bool, GLFW_MOUSE_BUTTON_LAST + 1> Input::mMouseButtons {};
+std::array<InputState, GLFW_KEY_LAST + 1> Input::mKeys {};
+std::array<InputState, GLFW_MOUSE_BUTTON_LAST + 1> Input::mMouseButtons {};
 // Initalize way outside the screen so it does not trigger anything before it should
 glm::vec3 Input::mMousePosition { std::numeric_limits<float>::max() };
 
-bool Input::IsKeyPressed(unsigned int keycode)
+void Input::Init(GLFWwindow* window)
 {
-    return mKeys[keycode];
+    for (auto& inputVal : mKeys)
+    {
+        inputVal = InputState::NotPressed;
+    }
+
+    for (auto& inputVal : mMouseButtons)
+    {
+        inputVal = InputState::NotPressed;
+    }
+
+    glfwSetKeyCallback(window, Input::KeyCallback);
+    glfwSetCursorPosCallback(window, Input::MousePositionCallback);
+    glfwSetMouseButtonCallback(window, Input::MouseButtonCallback);
+}
+
+void Input::Update()
+{
+    const auto updateInputState = [](InputState* inputStates, size_t size)
+    {
+        for (int i = 0; i < size; ++i)
+        {
+            if (inputStates[i] == InputState::Pressed)
+            {
+                inputStates[i] = InputState::Held;
+            }
+            else if (inputStates[i] == InputState::Released)
+            {
+                inputStates[i] = InputState::NotPressed;
+            }
+        }
+    };
+
+    updateInputState(mKeys.data(), mKeys.size());
+    updateInputState(mMouseButtons.data(), mMouseButtons.size());
 }
 
 void Input::KeyCallback(GLFWwindow* /*window*/, int key, int /*scancode*/, int action, int /*mods*/)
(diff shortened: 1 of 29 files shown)


engine/input.cpp
```diff
@@ -11,14 +11,47 @@ static constexpr int WINDOW_WIDTH = 1280;
...
+    }
+
+    glfwSetKeyCallback(window, Input::KeyCallback);
+    glfwSetCursorPosCallback(window, Input::MousePositionCallback);
+    glfwSetMouseButtonCallback(window, Input::MouseButtonCallback);
+}
+
+void Input::Update()
+{
+    const auto updateInputState = [](InputState* inputStates, size_t size)
```
jonmgomez: I don't suppose a good way to take these as a std::array is there? I looked into it, and it seems starting with C++20 you could do a templated lambda function. But there doesn't seem to be a good option otherwise
```
auto lambda = []<typename T>(T t){
};
```

cjappl: I think this may work? I know they added the ability to do auto in the params of lambdas in 17 or 14 or something...


```cpp

void Input::Update()
{
    const auto updateInputState = [](auto& inputStatesArray)
    {
        for (auto& state : inputStatesArray)
        {
            if (state == InputState::Pressed)
            {
                state = InputState::Held;
            }
            else if (state == InputState::Released)
            {
                state = InputState::NotPressed;
            }
        }
    };

    updateInputState(mKeys);
    updateInputState(mMouseButtons);
}
```

engine/uieventmanager.cpp
```diff
@@ -15,34 +17,66 @@ void UIEventManager::ProcessEvents(const UIElementCollection& uiElements)
...
     {
-        if (uiElement->GetType() != UIElementType::Button)
+        switch (uiElement->GetType())
         {
-            continue;
-        }
-
-        auto& button = static_cast<Button&>(*uiElement);
-        if (button.GetColliderBox()->CheckPointInBounds(mousePosition))
+        case UIElementType::Button:
```
cjappl: Ok, I'm seeing somewhat of an antipattern emerge here. 

Whenever you query an object for it's type, you should stop and think for a second. This is basically recreating inheritance to some degree.


So what can be done about it? There are a couple options, all with pros and cons:

1. Keep it as is

This is a very procedural coding style. The downside is that you have extremely tight coupling between the UIEventManager, and each UI Element type (if a UI element type changes, so must the event manager). Any time you add a new type, this ElementType enum needs to get updated, which triggers a recompilation of all UIElements, meaning there is also a too-tight coupling between these subtypes. At scale, this approach can get a little nasty.

2. Use inheritance

You could imagine each of your element types having a "ProcessEvent" overridden method. This would move all this logic into each class, so that each subclass has no idea about how the other subclasses handle this data (improving encapsulation). The downside of this is that you can see that over time this would bloat your UI element classes a little bit, as now they are not only UIElements but also event handlers.

In this case it will be easy to add new UiElement types, but more invasive to add other methods like this in the future (aka if you have to do another "global" action on all UiElements, you need to add that virtual method and all implementations to all subclasses)

3. Use the visitor design pattern

This is possibly the cleanest of the three, you could extend each UI Element such that it is amenable to Visitors. 

https://refactoring.guru/design-patterns/visitor

You create another class that is ProcessEventVisitor that has an override for each subtype. This removes the necessity for the "GetType" method, and keeps the object's guts inside of itself. This reduces the coupling between the UIEventManager and the UIElement classes.

This makes it easy to add new operations on all the types of UI elements, but harder to add new types (you have to add them to all your visitors).




In general, your approach here isn't the end of the world, but at scale this approaches downsides can become painful. You can see how it ballooned out of control at spatial with the event handlers and 1billion event types we had. I recommend at least considering as a thought exercise understanding the Visitor design pattern as another way to do this exact thing. It has pros and cons, but it would increase your knowledge of some fundamental approaches.


As for how to choose:
* If you have a fixed set of types, but an open set of operations on types - consider VISTIOR
* If you have a fixed set of operations, but an open set of types - consider INHERITANCE 
* If you want to tell me to go fuck myself, this works fine - consider THIS (which is fine, I don't take it personally :) )

gameobjects/settingsscreencontroller.cpp
```diff
@@ -0,0 +1,78 @@
...
+    mDifficultyText = Pong::AddUIElement<Text>("Difficulty", Config::GetValue<std::string>("font"), 1.0f, 75);
+    mDifficultyText->SetPosition(glm::vec3(-750.0f, 0.0f, 0.0f));
+
+    const Difficulty::Level currentDifficulty = Difficulty::GetLevel();
+    const std::array<std::string, 4> difficultyNames = { "Easy", "Medium", "Hard", "Impossible" };
+
+    for (int i = 0; i < 4; i++)
+    {
+        const float x = i % 2 * 550.0f;
+        const float y = (i < 2) * 150.0f;
```
cjappl: tehse calculations and many of these constants in this for block seem a little magical to me. Adding some named constants may be good.

cjappl (approved):

engine/input.cpp
jonmgomez: Yep worked like a charm. I didn't know you could declare params auto like that. Seems a bit weird to have that feature and also have templates. I'll need to look into it to understand that a bit more.

engine/uieventmanager.cpp
jonmgomez: Okay, I chose to try out the visitor pattern. I have been wanted to try out more design patterns because I don't really see many that often (In fact this is the first time I have seen or implemented visitor.) But yeah I like it. I think it will make adding and working with these ui elements in the future a lot nicer.

engine/processeventvisitor.h
```diff
@@ -0,0 +1,20 @@
...
+
+class Button;
+class CheckBox;
+class Slider;
+class Text;
+
+class ProcessEventVisitor
+{
+public:
+    virtual void VisitButton(Button& button) = 0;
```
cjappl: It would be more expected to have all of these just called "Visit", the type of the param makes it obvious what they are visiting

ui/button.cpp
```diff
@@ -17,9 +17,9 @@ void Button::Render() const
     mRectangle.Draw(mPosition);
 }
 
-UIElementType Button::GetType() const
+void Button::Accept(ProcessEventVisitor& uiElement)
 {
-    return UIElementType::Button;
+    uiElement.VisitButton(*this);
```
cjappl: awesome.

ui/uielement.h
```diff
@@ -27,20 +28,12 @@ struct MeshComponent
     }
 };
 
-enum class UIElementType
-{
-    Button,
-    CheckBox,
-    Slider,
-    Text
-};
```
cjappl, line 17: lovely

engine/input.cpp
cjappl: Yeah, I actually think these are templates under the hood, like if you look at what the compiler generates it's just a template. Funny they took this approach first...

engine/uieventmanager.cpp
cjappl: Awesome. Good works, looks like I would expect. If you want to chat over the implications of this or any other design pattern, I'm reading a very interesting book on them right now and would love to converse about them to deepen my own knowledge. I also see Design Patterns as a weakness in my coding skills.

The book is [here](https://www.amazon.com/Software-Design-Principles-Patterns-High-Quality/dp/1098113160), and so far it is SUPER useful, finally helping them click but it is dense.

cjappl: (I have also been surprised at how many design patterns I had encountered but just didn't know the proper names, so you know a lot of them, you just don't know it yet ;) )

cjappl (approved):

PR #16, opened 2023-12-08
Create classes for containing application window, and game engine.

jonmgomez:
Not a ton of new code. A lot of this is moving around code and dealing with class structure. 
Previously many of these systems were in main.cpp and are now into, ApplicationWindow which is meant to manage the actual window for whatever platform, and the Engine class which is meant to encapuslate the entire game "engine" as a whole. This also introduces the ability to dynamically resize the window without breaking the game. Before it would have been much more annoying to do.

### DIFF
--- a/engine/input.cpp
+++ b/engine/input.cpp
@@ -64,41 +64,19 @@ void Input::KeyCallback(GLFWwindow* /*window*/, int key, int /*scancode*/, int a
     mKeys[key] = (action == GLFW_PRESS || action == GLFW_REPEAT) ? InputState::Pressed : InputState::Released;
 }
 
-bool Input::CheckKeyDown(unsigned int button)
+InputState Input::GetKeyState(unsigned int keycode)
 {
-    return mKeys[button] == InputState::Pressed || mKeys[button] == InputState::Held;
+    return mKeys[keycode];
 }
 
-bool Input::CheckKeyUp(unsigned int button)
+bool Input::CheckKeyDown(unsigned int keycode)
 {
-    return mKeys[button] == InputState::Released || mKeys[button] == InputState::NotPressed;
+    return mKeys[keycode] == InputState::Pressed || mKeys[keycode] == InputState::Held;
 }
 
-// Mouse position converted to screen/world coordinates
-glm::vec3 Input::GetMousePosition()
+bool Input::CheckKeyUp(unsigned int keycode)
 {
-    return mMousePosition;
-}
-
-void Input::MousePositionCallback(GLFWwindow* /*window*/, double xPos, double yPos)
-{
-    std::cout << "Raw Mouse position: " << xPos << ", " << yPos << std::endl;
-    const float x = static_cast<float>(xPos);
-    const float y = static_cast<float>(yPos);
-
-    const int windowWidth = ApplicationWindow::GetScreenWidth();
-    const int windowHeight = ApplicationWindow::GetScreenHeight();
-    const float halfWindowWidth = windowWidth / 2.0f;
-    const float halfWindowHeight = windowHeight / 2.0f;
-
-    const float mouseWindowX = (x < halfWindowWidth) ?
-        (halfWindowWidth - x) * -2.0f : (x - halfWindowWidth) * 2.0f;
-    // y coords are positive going down, so we need to invert them
-    const float mouseWindowY = (y <= halfWindowHeight) ?
-        (halfWindowHeight - y) * 2.0f : (y - halfWindowHeight) * -2.0f;
-
-    mMousePosition.x = mouseWindowX / windowWidth * Renderer::GetXCoordMax();
-    mMousePosition.y = mouseWindowY / windowHeight * Renderer::GetYCoordMax();
+    return mKeys[keycode] == InputState::Released || mKeys[keycode] == InputState::NotPressed;
 }
 
 void Input::MouseButtonCallback(GLFWwindow* /*window*/, int button, int action, int /*mods*/)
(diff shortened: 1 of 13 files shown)


cjappl:
If this is a lot of moving around stuff, please comment where you want me to look specifically so I don't review things that were just shuffled :)

jonmgomez:
> If this is a lot of moving around stuff, please comment where you want me to look specifically so I don't review things that were just shuffled :)

Yeah, good point. I think mostly I am looking at how the structure of this looks. I made all of these classes singletons, but I don't think they all have to be. However, I did this because I want the ability to use some of the calls to these classes as a namespace essentially. So instead of some call to `Quit()` or `mEngine.Quit()` it refers to just `Engine::Quit()` where engine appears to be a namespace more than a class itself. Same thing with `ApplicationWindow::Resize()` or `Input::GetMousePosition()` Wondering if that approach is feasable long term or if it may introduce some issues.

Additionally, I think a future PR may involve separating the rendering and the game logic into separate threads. That will make the threading a lot more complex. I was curious if it still makes sense to use these classes statically. I know static functions and variables can get a bit more complicated when threads are introduced.

cjappl:
So one of the major downsides of singletons, especially global static singletons is that they hide complexity and dependencies.

This makes things harder to change, and harder to test. In general you should have as few singletons as humanly possible, and I could see a case for having no singletons at all (this is a new belief of mine, I haven't tried it in practice). Singletons are SO EASY to use, and that's why they are so tempting. 

But, imagine you're writing a test for some function that uses a global static singleton. You then have to write some test code like:

```
CreateSingleton1()
CreateSingleton2()

doThingUnderTest();
```

there is no indication that doThingUnderTest needs these two singletons to work properly, you probably just crashed when you ran and learned the lesson.

Instead, you could consider explicit dependency injection, where you pass in everything your class or method depends on:

```
logger = MakeLogger();
randomNumberGenerator = MakeRandomNumberGenerator();

doThing(logger, randomNumberGenerator);

```

Uglier function signature, but it does show that these are EXPLICITLY dependent on one another. There are pros and cons, but knowing what modules depend on each other can make your life easier over time.


In terms of "Namespacing" functions, you can either use static methods, or just put free functions in an associated namespace that has the same or similar name as your class. I would not use a singleton just to get namespaceing out of it.

engine/input.cpp
```diff
@@ -28,9 +28,10 @@ void Input::Init(GLFWwindow* window)
         inputVal = InputState::NotPressed;
     }
 
-    glfwSetKeyCallback(window, Input::KeyCallback);
-    glfwSetCursorPosCallback(window, Input::MousePositionCallback);
-    glfwSetMouseButtonCallback(window, Input::MouseButtonCallback);
+    GLFWwindow* glfwWindow = ApplicationWindow::GetInstance().GetWindow();
```
cjappl, line 31: This definitely hides this dependency a little. If you ever try to unit test this, or change the implementation it will be messy to tear it apart again.

cjappl (approved):
Overall, there are pros and cons to having a heavily singleton laden codebase. It's up to you to decide if it is right for your usecase!

I have previously been a fan for bigger top level modules (like a logger), but more recently I've begun thinking it may be a mistake overall because it hides dependencies and seems like a cheap way to cut around careful architecture.

This is an open line of thinking, so in total ¯\_(ツ)_/¯

jonmgomez:
> So one of the major downsides of singletons, especially global static singletons is that they hide complexity and dependencies.

I think I may keep it around for the moment while I tinker with other designs. Although it does bring a up another question. These singletons are a little hard to test already (at least unit test) because they require the main engine/game loop in order to function properly. How would you be able to test these nicely? Is it only really feaseable to have some kind of automated testing system like we had at Spatial?

cjappl:
> These singletons are a little hard to test already (at least unit test) because they require the main engine/game loop in order to function properly. How would you be able to test these nicely? Is it only really feaseable to have some kind of automated testing system like we had at Spatial?

So a couple things. 

1. There are different levels of testing (which I'm sure you're generally aware of)

there is everything from unit testing (one function) to system testing (like python at spatial). In between you have integration testing, and probably some other's I am missing.

A good testing coverage has all of these leves. So ideally you'd have some unit, integration and system testing to minimize the chance that manual QA catches something. I have yet to work in a project where manual QA can be totally omitted, I think that is a pipe dream...

2. Some things are reasonable to unit test, other things not.

Things like hardware interaction, logging modules etc are very hard to test, and aren't really worth the effort. I think a big part of development is understanding what is/isn't worth the effort to test.

Another thing is designing for test-ability needs to happen almost immediately, because designing for some other usecases (speed, ease of use) can conflict with this. Sometimes boltin on testing after the fact can be difficult. This is why we had pretty decent unit testing of "leaf node classes" at spatial, but absolutely dogshit testing of any modules even one level higher. We just did not design for testability which ended up in a lot of churn and bugs. Eventually lack of testing becomes like pouring concrete over your codebase, it makes any change difficult and bugprone.


That being said, some things that make testing easier:
* Ensuring every class and function has one responsibility, and one responsibility only
* Dependency injection, where you don't hide dependencies away inside classes, but inject them from the outside. This can make mocking functionality easier
* (as stated above) knowing what is going to have diminishing returns testing wise, and what is critical. The things that are critical to test are the things that are most likely to change in the future (does my new graphics library work as well as the old one? If I change this datastructure from a vector to a map will it break?) as well as the things that have high dependencies on other modules.

On the dependencies front: You remember how in Spatial, you changed something in one module way across the codebase and some other piece of the code would fall over? That's an example of both having too many dependencies that weren't understood, and not having tests for them. The solution there is:

1. Reduce concrete dependencies
2. Where necessary, test the hell out of them.


As for specific tests for your specific classes, if you give me an example of a class and a test that is hard to write I can give you a more concrete answer :) hope this rambling was helpful.

jonmgomez:
> As for specific tests for your specific classes, if you give me an example of a class and a test that is hard to write I can give you a more concrete answer :) hope this rambling was helpful.

Yeah, so there are quite a few I was thinking of that you could probably chalk up to hardware interactions. Essentially anything with opengl or glfw falls into that. Listening for keyboard callbacks and handling their state, managing the window, and any of the opengl classes (renderer, shader, vertex buffer, and so on...)

cjappl:
Yeah, there is always going to be a level that is essentially untestable in any form, which covers the things you are talking about.

It's your job as a developer to make that layer as thin as humanly possible, so it's feasible that if your 3rd party package is working right, your code works.


Take for instance writing a Keyboard input class directly the the OS system call level. You could have a lot of very library specific calls scattered throughout your code, or you could write to a known interface that wrapped all of that stuff and pushed it very far down.

If you wrote your other  classes in a way that they weren't dependent at all on the implementation details, you could swap in a "TestKeyboard" for those classes and write unit tests against them. Your `Keyboard` class again is untestable, but if you encapsulate it well enough you can test all the classes it interacts with without too much pain.

In terms of design patterns, and concepts, check out Adapter, Strategy, Command as well as the SOLID principles. In particular the D stands for "Dependency Inversion" and is kind of what I'm getting at here.

PR #17, opened 2023-12-18
Introduce Entity Component System

jonmgomez:
A pattern that is in most game engines (Unity, Unreal, Godot) is the Entity Component System. This pattern allows for the creation of unique gameobject/entities that have very specific behaviors, but may share behaviors between each other through the use of components. Instead of using inheritance to create game objects with meshes/colliders/transforms, this uses composition to attach all components to this game object. 

Up until now the gameobject class had a mesh and collider as a protected variable, so any gameobjects like the player can create both the mesh and collider, while the score controller only creates a collider. Instead these classes override a `InitalizeComponents()` function where they `AddComponent<Component>()` with each component needed. The system part are like the collision manager. These systems operate on the relevant components.

Commonly when using an ecs is trying to optimize for cache locality. When iterating over thousands of gameobjects, if the components are stored contiguously that can be significantly better than stored around. So instead of storing the components on the entity/gameobject, they are primarily allocated in a vector for each component.

I think there may be multiple parts to this. Right now, gameobjects hold components, while also supporting the `OnStart` and `OnUpdate` functions and behavior. Instead this should be a component itself and be attached to a purely empty gameobject.


engine/component.h
```diff
@@ -0,0 +1,73 @@
...
+    {
+        mComponents.push_back(std::move(component));
+    }
+
+    static const std::vector<std::unique_ptr<T>>& GetComponents()
+    {
+        return mComponents;
+    }
+
+    static std::vector<std::unique_ptr<T>> mComponents;
```
jonmgomez: Is there any other way to do this sort of thing? Essentially, with an unknown number of component types, it needs to create a vector for each component type even if it is user defined.

engine/colliderbox.h
```diff
@@ -1,13 +1,15 @@
...
+
 #include <glm/glm.hpp>
 
 #include <array>
 
 namespace pong
 {
 
-class ColliderBox
+class ColliderBox : public Component<ColliderBox>
```
cjappl, line 12: Ooh, the CRTP!! Literally was just reading a chapter of a book about it.

This was also how Roblox did static compile time polymorphism, so you're in good company "game engine" wise.

engine/collisionmanager.cpp
```diff
@@ -9,16 +10,26 @@
...
-void CollisionManager::ProcessCollisions(const std::vector<std::unique_ptr<GameObject>>& gameObjects)
+void CollisionManager::ProcessCollisions(const std::vector<std::unique_ptr<ColliderBox>>& colliders)
 {
-    for (int index = 0; index < gameObjects.size(); index++)
+    for (int index = 0; index < colliders.size(); index++)
     {
-        auto& gameObject = gameObjects[index];
-        for (int otherIndex = index + 1; otherIndex < gameObjects.size(); otherIndex++)
+        auto& collider = colliders[index];
+        if (collider->GetGameObject()->GetComponent<Transform>() != nullptr)
```
cjappl: I would consider using an initializer in the if statement here because you use this so frequently in the block


I believe the syntax is something like
```
if (auto* transform = collider->GetGameObject()->GetComponent<Transform>(); transform != nullptr)
{
                collider->OnPositionUpdate(transform->mPosition);
...    
}
```

engine/component.h
cjappl: No, I think this is fine I think. I can't think of another way to do it.

engine/component.h
```diff
@@ -0,0 +1,73 @@
...
+
+template<typename T> int TypeIDGenerator<T>::sID = 0;
+
+template<typename ComponentSubType>
+int GetComponentTypeID()
+{
+    return TypeIDGenerator<BaseComponent>::GetID<ComponentSubType>();
+}
+
+template<class T>
```
cjappl: Often when using the CRTP this template parameter is called "Derived"

gameobjects/ball.cpp
```diff
@@ -20,13 +20,22 @@ static constexpr float Y_STARTING_POSITION_BOUNDS = 500.0f;
 
 static constexpr std::chrono::seconds BALL_RESET_WAIT_S { 3 };
 
+void Ball::InitalizeComponents()
+{
+    AddComponent<Transform>(glm::vec3(0.0f));
+    AddComponent<Rectangle>(BALL_WIDTH, BALL_WIDTH);
+    AddComponent<ColliderBox>(BALL_WIDTH, BALL_WIDTH);
```
cjappl, line 27: Cool. I really like this for a lot of reasons.

You see how this approach naturally separates concerns and interfaces into individual sub classes. Now when you want to change something that "is transformable" you know where to look. You don't have a base class doing it, you don't have to update it in a billion subclasses.


This is really mature architecture, very thought provoking.

engine/pong.cpp
```diff
@@ -136,24 +139,29 @@ void Pong::LoadScene(Scene scene)
             constexpr float verticalWallX = 1280;
 
             auto topWall = std::make_unique<Wall>(horizontalWallWidth, horizontalWallHeight);
-            topWall->SetPosition(glm::vec3(0.0f, horizontalWallY, 0.0f));
+            topWall->InitalizeComponents();
```
cjappl, line 142: What is the advantage of having InitializeComponents externally here, instead of in the constructor (which will be naturally called on line 141).

With this being external it seems easy for someone else (or you) to mix it up and not call it in a new copy/paste

gameobjects/gameobject.h
```diff
@@ -20,40 +21,60 @@ class GameObject
...
+            return static_cast<T*>(mComponents[componentID]);
+        }
+
+        return nullptr;
+    }
+
+    template<typename T, typename... Args>
+    T* AddComponent(Args&&... args)
+    {
+        std::unique_ptr<T> component = std::make_unique<T>(std::forward<Args>(args)...);
```
cjappl: now we are modern c++ing. This is an advanced ass statement.

`auto` could be used here, unique_ptr is obvious

gameobjects/scorearea.cpp
```diff
@@ -11,13 +11,21 @@
 namespace pong
 {
 
-ScoreArea::ScoreArea(float width, float height, bool playerSide)
+ScoreArea::ScoreArea(float width, float height, bool playerSide) :
+    mWidth { width },
+    mHeight { height }
 {
-    mColliderBox = std::make_unique<ColliderBox>(width, height);
     mIsPlayerScoreArea = playerSide;
```
cjappl: mIsPlayerScoreArea = playerSide;

could be moved into the initializer list with the others

cjappl (approved):
Overall I think this change is a really big improvement! I learned a thing or two reading it. Very clever system.

Do you have any resources of where you read/learned about it?

engine/pong.cpp
jonmgomez: Hmm well since it is a pure virtual it will still need to manually be added to the constructor right? I figured it would kind of be the same in that it could get missed when creating a new object. 
At least for `OnStart()` it needs to be called externally because things won't be ready properly unless the object is first positioned and added to the gameobjects list. That may not be the case now for `InitalizeComponents`, but I imagine it could be sometime in the future? 

Eventually I did want to clean up this part of the code so that you don't need to copy paste the whole.
`std::make_unique<>`
`InitalizeComponents()`
`GetComponent<Transform>()->mPosition =`
block, with a function to do each time.

cjappl: Cool. Put that down as a TODO item and we will re-discuss it there. I think figuring out how to do that would be valuable

jonmgomez:
> Do you have any resources of where you read/learned about it?

Yeah a couple pages I was learning from and taking inspirations from
https://devlog.hexops.com/2022/lets-build-ecs-part-1/
https://austinmorlan.com/posts/entity_component_system/
And this guy on youtube gave me a really good overview of the idea:
https://www.youtube.com/watch?v=Z-CILn2w9K0

gameobjects/ball.cpp
jonmgomez: Yeah it is definitely a lot nicer 🎉

PR #18, opened 2023-12-20
Entity Component System: Behavior components and better scene creation (Part 2)

jonmgomez:
With the entity component system (ecs) introduced, gameobjects now are composed of mulitple components to give them behavior. However the idea of a game object having behavior through `OnStart()`, `OnUpdate()`, or `OnCollision()` callbacks was tied to gameobjects themselves. For a proper ecs, this should be in its own component `Behavior.` It is a little bit trickier though because Behaviors are meant to be overridden by users. And overriding the base Behavior class causes problems with the CRTP Component chain. Instead of different behavior types being stored in their own vector, they are stored in the Behavior vector. So we need separate typing to be able to get out certain behaviors out from the others.

Another big change here is how scenes are loaded. For now, a Scene class will define a function to `BuildScene()` which creates the scene from helper functions `NewObject()` and `NewObjectFromBlueprint()`. An object Blueprint is nothing more than a way to quickly define multiple instances of the same object. These objects are put into a vector and handed of to the game loop to be properly loaded.

### DIFF
--- a/engine/behavior.cpp
+++ b/engine/behavior.cpp
@@ -6,7 +6,7 @@
 namespace pong
 {
 
-void Behavior::SetTimeout(std::chrono::duration<double> timeout, std::function<void()> callback)
+void Behavior::SetTimeout(const std::chrono::duration<double>& timeout, const std::function<void()>& callback)
 {
     Pong::GetInstance().GetTimer().AddTimer(GetGameObjectID(), timeout, callback);
 }
--- a/engine/behavior.cpp
+++ b/engine/behavior.cpp
@@ -1,0 +1,24 @@
+#include "behavior.h"
+
+#include "audiomixer.h"
+#include "pong.h"
+
+namespace pong
+{
+
+void Behavior::SetTimeout(std::chrono::duration<double> timeout, std::function<void()> callback)
+{
+    Pong::SetTimeout(GetGameObjectID(), timeout, callback);
+}
+
+void Behavior::PlaySound(const Sound& sound)
+{
+    Pong::GetInstance().GetAudioMixer().PlaySound(sound);
+}
+
+void Behavior::PlaySound(const Sound& sound, const glm::vec3& position)
+{
+    Pong::GetInstance().GetAudioMixer().PlaySound(sound, position);
+}
+
+} // namespace pong
--- a/engine/pong.cpp
+++ b/engine/pong.cpp
@@ -1,6 +1,7 @@
 #include "pong.h"
 
 #include "ball.h"
+#include "behavior.h"
 #include "colliderbox.h"
 #include "config.h"
 #include "input.h"
@@ -13,6 +14,7 @@
 #include "settingsscreencontroller.h"
 #include "titlescreencontroller.h"
 #include "wall.h"
+#include "pongscenes.h"
 
 namespace pong
 {
@@ -47,7 +49,7 @@ void Pong::Init()
 {
     GetInstance().GetAudioMixer().Init();
     GetInstance().GetTimer().Init();
-    GetInstance().LoadScene(Scene::TitleScreen);
+    GetInstance().LoadScene(SceneType::Game);
 }
 
 void Pong::GameLoop()
@@ -62,9 +64,9 @@ void Pong::GameLoop()
         GetInstance().mChangeSceneRequested = false;
     }
 
-    for (auto& gameObject : GetInstance().mGameObjects)
+    for (auto& behavior : Behavior::GetComponents())
     {
-        gameObject->OnUpdate();
+        behavior->OnUpdate();
     }
 
     Renderer::DrawAllMeshes(Mesh::GetComponents());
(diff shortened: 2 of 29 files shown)


engine/pong.h
```diff
@@ -36,19 +36,38 @@ class Pong
...
+    // Returns the first component of type T found in the scene
+    template<typename T, typename = std::enable_if_t<std::is_base_of_v<BaseComponent, T> && !std::is_base_of_v<Behavior, T>>>
+    static T* FindComponentOfType()
     {
-        for (auto& gameObject : GetInstance().mGameObjects)
+        const std::vector<std::unique_ptr<T>>& components = T::GetComponents();
+
+        if (components.empty())
         {
-            auto castedGameObject = dynamic_cast<T*>(gameObject.get());
```
jonmgomez, line 46: Moving to the ECS and using CRTP also solved these dynamic casting issues. Very cool! 🎉

engine/pongscenes.h
```diff
@@ -0,0 +1,31 @@
...
+#include <memory>
+#include <vector>
+
+namespace pong
+{
+
+class TitleScene : public Scene
+{
+public:
+    void BuildScene() override;
```
jonmgomez, line 15: Curious if you have any thoughts on how I switched up the scene creation. I think it saves from some copy pasting, but still kind of follows the same idea of building a vector of objects. 

A proper way this is handled in game engines is with .scene asset. These consist of a lot of object data (json/yaml) that define the layout of a scene. But I think that will be another challenge converting text into specific component types 😬.

gameobjects/scorearea.cpp
```diff
@@ -11,24 +11,31 @@
...
 {
 
-ScoreArea::ScoreArea(float width, float height, bool playerSide) :
+ScoreAreaBlueprint::ScoreAreaBlueprint(float width, float height, bool playerSide) :
     mWidth { width },
     mHeight { height },
     mIsPlayerScoreArea { playerSide }
 {
     SetInstanceName("ScoreArea");
+    InitalizeComponents();
```
jonmgomez: So I did opt to just call `InitalizeComponents()` from within the constructor. However I am curious if is it even worth doing it this way instead of calls to `AddComponent<>()` in the constructor itself. With this current method it forces `InitalizeComponents` to be called and requires saving the parmeters from the constructor for the creation of the components later on.

engine/behavior.cpp
```diff
@@ -0,0 +1,24 @@
+#include "behavior.h"
+
+#include "audiomixer.h"
+#include "pong.h"
+
+namespace pong
+{
+
+void Behavior::SetTimeout(std::chrono::duration<double> timeout, std::function<void()> callback)
```
cjappl: Do these need to be const ref?

engine/pong.cpp
```diff
@@ -92,85 +99,43 @@ void Pong::Cleanup()
...
-    switch (scene)
+    const auto LoadSceneObjects = [&](auto scene) {
+        scene.BuildScene();
+        GetInstance().mGameObjects = scene.TransferGameObjects();
+    };
+
+    switch (sceneType)
     {
-        case Scene::TitleScreen:
+        case SceneType::Title:
```
cjappl, line 119: Another use for the Visitor perhaps?

engine/pongscenes.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "settingsscreencontroller.h"
+#include "titlescreencontroller.h"
+#include "wall.h"
+
+namespace pong
+{
+
+void TitleScene::BuildScene()
+{
+    GameObject* titleScreenController = NewObject();
```
cjappl, line 17: Does this allocate? looks like possibly a leak could be happening here

engine/pongscenes.h
cjappl: I think this is good. I think this is basically the factory design pattern, you could look into that if you wanted some inspiration on a design like this

https://refactoring.guru/design-patterns/factory-method
https://refactoring.guru/design-patterns/abstract-factory

engine/scene.h
```diff
@@ -0,0 +1,55 @@
...
+using GameObjectCollection = std::vector<std::unique_ptr<GameObject>>;
+
+class Scene
+{
+public:
+    virtual void BuildScene() = 0;
+
+    GameObjectCollection TransferGameObjects()
+    {
+        return std::move(mGameObjects);
```
cjappl, line 25: Don't return and move, it ruins copy-elision and is often slower.

cjappl: (just return directly, it does the optimal job magically)

gameobjects/gameobject.cpp
```diff
@@ -44,19 +26,29 @@ void GameObject::SetInstanceName(const std::string& name)
...
+
+    for (auto& behavior : allBehaviors)
+    {
+        if (behavior->GetGameObjectID() == mId)
+        {
+            behaviors.push_back(behavior.get());
+        }
+    }
+
+    return behaviors;
```
cjappl, line 43: This is a vector of non-owning behaviors. Be careful about their lifetimes and the person requesting them not using them after they die! 

There isn't a bug here, but it is bug-possible

gameobjects/opponent.h
```diff
@@ -10,10 +11,17 @@ namespace pong
...
 
-class Opponent : public GameObject
+class OpponentBlueprint : public GameObjectBlueprint
 {
 public:
+    OpponentBlueprint();
     void InitalizeComponents() override;
+};
+
+class Opponent : public Behavior
```
cjappl, line 20: This seems like possibly a bad inheritance hierarchy. Does it make sense that an Opponent "is-a" behavior?

it seems like an opponent may "have-a" behavior.

gameobjects/scorearea.cpp
cjappl: Good question. I think I would likely just flatten it together in the Constructor. I see what you're trying to do by forcing all your subclasses to initialize, but I think it is just boilerplate that doesn't actually prevent a bug.

I'll ponder if there is any way to enforce what you want to enforce, nothing is coming to mind right now...

cjappl (approved):

engine/pongscenes.cpp
jonmgomez: Yeah it does allocate. The idea I had for these `BuildScene` functions would be that the scene would be created using helper functions to create objects and add them to the scene's internal object list. It would also return a pointer incase anything also needed to be done to the object, whether setting their position, name, or setting some Behavior member variable. Only after this Build function then the gameobject list will be moved and given to the main game loop to be properly used.

engine/pong.cpp
jonmgomez: Yeah thats valid. I think I will keep if for now just because it is a really simple system, and I don't plan on adding more than maybe one more scene type.

engine/scene.h
jonmgomez: So I tested this out and it comes up with some compiler errors for trying to copy a vector of `unique_ptr`s. I assumed that because this is returning a vector from a class member variable it would not implicitly move it? So it must be done explicitly.

gameobjects/gameobject.cpp
jonmgomez: Yeah I have thought about this previously. I think it comes down to the idea that nothing should ever be deleted under the user without them explicitly asking for it. So if two objects collide and one is destroyed, it is on the user to know that it should stop using that object after that point. I have done some things in Unity where certain objects being destroy would callback relevant objects before destruction so they can remove the object from various lists and variables. 

It is nicer in Unity using C# though because any variables pointing to a now deleted object get automatically `null`'d without needed to set it to `null`. So instead of doing callbacks, you can also just put `if (object != null)` around everything and it would work the same

gameobjects/opponent.h
jonmgomez: Yeah I understand that. It may just be a game engine naming thing. In Unity the script components are called "MonoBehavior." Generally I have always named things in similar style. The entire player entity/object may have colliders, meshes, lights, multiple monobehaviors, but then also has the Player MonoBehavior to control movement and other things exclusive to the player. It could be renamed to something like PlayerScript or PlayerBehavior, but I think it generally gets the idea across that it is the main controller of the object's behavior.

engine/pongscenes.cpp
cjappl: I think this is good, fits in line with "raw pointers being non-owning.

engine/scene.h
cjappl: oh! Wow, interesting use case. cool, Today I Learned

gameobjects/opponent.h
cjappl: Cool!! Works for me, just making sure you have considered it, which seems like it's well thought out :)

PR #19, opened 2023-12-31
Add Scene deserialization from json .scene files

jonmgomez:
Previously, to load a scene within the game, it was done by building it using code to create each object and component consecutively:
```
GameObject* object = NewObject();
object->AddComponent<Transform>(glm::vec3(0.0f));
object->AddComponent<Mesh>(std::make_unique<Rectangle>(100.0f, 100.0f));
// and so on...
```

Instead, most game engines have a way to serialize a scene and deserailze it from some format. For example Unity loads scene files from a YAML format. This allows for changing the starting layout of a scene without recompiling the project.

A lot of stuff can be ignored here. Some of the component (slider, checkbox, button) needed to have a default constructor in order to work with the deserializer. So now they can be default constructed, then have properties set on them, and then recalculate necessary variables. This also removes old scene loading using pongscenes.h/.cpp and scenes.h

**Main files to look at would be sceneloader.h/.cpp and componentdeserializer.h/.cpp**


engine/scene/componentdeserializer.h
```diff
@@ -0,0 +1,52 @@
...
+    void VisitComponent(ColliderBox* component);
+    void VisitComponent(Mesh* component);
+    void VisitComponent(Slider* component);
+    void VisitComponent(Text* component);
+    void VisitComponent(Transform* component);
+
+
+    void VisitComponent(Behavior* component);
+    // User defined components
+    void VisitComponent(Opponent* component);
```
jonmgomez, line 42: So using the visitor pattern again becuase I think it works the nicest. I did want to ask if you can think of any good ways to do this kind of deserialization of custom and user defined components automatically? For right now, if a new behavior component is created it requires adding a new function to handle it here, as well as mapping any properties that should be setting in the scene file.

ie. A Player component in a .scene file with the ability to set its starting `mSpeed` variable needs a mapping from getting the json value `["speed"]` to calling `SetSpeed()` on the component.

In Unity this kind of thing is done directly in the code by marking variables with [SerializeField] which make it setable from the engine inspector, and from scene creation.

I think it is a very tricky problem. It seems a little bit more clear with unity since it interprets C#.

engine/scene/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+set(SOURCES
+    sceneloader.cpp
```
jonmgomez, line 2: Ah the return of the sceneloadmanager! 🔥

engine/scene/componentdeserializer.h
```diff
@@ -0,0 +1,52 @@
...
+class Mesh;
+class Player;
+class Opponent;
+class ScoreArea;
+class ScoreController;
+class SettingsScreenController;
+class Slider;
+class Text;
+class TitleScreenController;
+class Transform;
```
jonmgomez, line 23: Seems like the visitor pattern is a little prone to inviting circular dependencies. Is there a better way to fix these? It is not the worst to forward declaring everything, but it is less clean.

engine/scene/componentdeserializer.cpp
```diff
@@ -0,0 +1,162 @@
...
+    component->SetSpeed(speed);
+}
+
+void ComponentDeserializer::VisitComponent(ScoreArea* component)
+{
+    const bool playerSide = mCurrentJson["player_side"];
+    component->SetIsPlayerSide(playerSide);
+}
+
+void ComponentDeserializer::VisitComponent(ScoreController* /*component*/)
```
cjappl, line 157: Nice use of Visitor in this file.

What are your reflections after using it more heavily? What are the pros and cons of using it? would you use it again?

engine/scene/componentdeserializer.h
cjappl: Yep, this version is known as the "cyclic visitor" and is specifically known for needing a circular reference. There is an "acylic visitor" but the performance cost comes into play with another direction.

There is a way around it using value semantics std::variant and std::visit, but it is a bit of a change.

He talks about it in detail here if you are curious:
https://youtu.be/PEcy1vYHb8A?si=y2rM3EAlMZUu8Aju



I think the forward declaration here is good. Not that it prevents things from being recompiled if something changes in the Opponent.h header.

engine/scene/componentdeserializer.h
cjappl: I'm not sure if there is a great way to do it "automatically" as in "no code". This is possible due to the lack of C++ "Reflection" which is something I believe C# has. That's a way to introspect fields and methods of an object like you're describing. (Apparently this is coming in c++26 or 29 and will be pretty insanely revolutionary for the language).



I think any solution that would work today requires some further coupling between the components. One benefit of the visitor pattern is that the responsibility of "Serialization" is all in this component. Each `Opponent` or similar has no concept at all of anything regarding serialization.

Any solution (for instance, providing a virtual method to return a `std::map` of keys to functions to call) injects some knowledge of serialization into the `Component` class itself.


Overall, I think it's a benefit to keep these concerns separate. It can help keep things divided in your mind and makes changes easier in the future.

engine/scene/sceneloader.cpp
```diff
@@ -0,0 +1,198 @@
...
+#include <chrono>
+#include <fstream>
+
+namespace pong
+{
+
+SceneLoader::SceneLoader()
+{
+    // Engine components
+    mComponentMappings["Transform"] = [](GameObject* gameObject) {
```
cjappl: Lots of duplication in this function. Is it worth it to abstract these lambdas to some template function? Or do you expect them to be different over time?

engine/scene/sceneloader.cpp
```diff
@@ -0,0 +1,198 @@
...
+    };
+
+    mComponentMappings["SettingsScreenController"] = [](GameObject* gameObject) {
+        return gameObject->AddComponent<SettingsScreenController>();
+    };
+}
+
+void SceneLoader::PreLoadScenes()
+{
+    std::vector<std::string> sceneNames = Config::GetValue("scenes", std::vector<std::string> {});
```
cjappl, line 59: is the second type here necessary? I am wondering if CTAD is smart enough to do it automagically.

```
std::vector<std::string> sceneNames = Config::GetValue("scenes", {});
```

(Total curiosity question, I don't know the answer)

engine/scene/sceneloader.cpp
```diff
@@ -0,0 +1,198 @@
...
+        return {};
+    }
+
+    const nlohmann::json& sceneJson = mScenes[sceneName];
+    return LoadSceneFromJson(sceneJson);
+}
+
+std::vector<std::unique_ptr<GameObject>> SceneLoader::LoadSceneFromJson(const nlohmann::json& sceneJson)
+{
+    std::vector<std::unique_ptr<GameObject>> gameObjects {};
```
cjappl: You use this type enough that it may be worth abstracting to a well named `using`

engine/scene/sceneloader.cpp
```diff
@@ -0,0 +1,198 @@
...
+            componentDeserializer.DeserializeComponent(newComponent, componentJson);
+        }
+
+        gameObjects.push_back(std::move(gameObject));
+    }
+
+    const auto endTime = std::chrono::high_resolution_clock::now();
+    const auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(endTime - startTime).count();
+
+    LogInfo("Scene Loaded, Time Taken: {}ms ({}s). Total Objects: {} Total Components: {}",
```
cjappl, line 158: Cool!! I don't know if I missed this or forgot it over the holidays, but using `fmtlib` or something similar. I've really loved it when I've used it in the past.

No more remembering what the hell the format specifier for `unsigned size_t`

engine/scene/sceneloader.h
```diff
@@ -0,0 +1,34 @@
...
+    SceneLoader();
+    ~SceneLoader() = default;
+
+    void PreLoadScenes();
+    std::vector<std::unique_ptr<GameObject>> LoadScene(const std::string& sceneName);
+
+private:
+    std::vector<std::unique_ptr<GameObject>> LoadSceneFromJson(const nlohmann::json& sceneJson);
+
+    std::unordered_map<std::string, nlohmann::json> mScenes {};
```
cjappl, line 31: One thing to note here is that you're gluing yourself to `nlohmann::json`.

Any time a tps shows up in a header, it means that there is a transitive physical dependency on this tps, and it "leaks" into the rest of your codebase.

e.g. if `A.h` has `json` dependency, `B.h` includes `A.h`, any `C.h` that includes `B.h` then needs to know and care about `json.


There are ways to get around this, whicih we can talk about, but it may not be worth it in the short run. Just trying to get you to think about dependencies and how they restrict future changes. Some of the best designed codebases are designed primarily to be Easy To Change as a primary intention.

Would changing your dependency to `nlohmann::json` be easy or hard? What would necessitate a change like that? Is it worth it to design around a possible change like that? (architectural questions to ask yourself)

gameobjects/settingsscreencontroller.cpp
```diff
@@ -14,17 +14,22 @@ void SettingsScreenController::OnStart()
 {
     mBackButton = Pong::FindGameObjectByName("BackButton")->GetComponent<Button>();
     mBackButton->AddListener(ButtonEvent::Pressed, []() {
-        Pong::LoadSceneNext(SceneType::Title);
+        Pong::LoadSceneNext("Title");
```
cjappl, line 17: Any reason you switched from an enum to a string here? (no preference, just curious)

cjappl (approved):
Nice!

engine/scene/sceneloader.cpp
jonmgomez: Just to make sure I know what you mean some thing like 
`mComponentMappings["Transform"] = AddComponentToObject<Transform>(object);`?
I think that would make sense, I don't really expect it to change much.

gameobjects/settingsscreencontroller.cpp
jonmgomez: Yeah, this way you can load the scene by the name in the json file instead of hardcoding an enum for each scene.

engine/scene/sceneloader.h
jonmgomez: Yeah that is a good point. Now that I have worked with C++ a lot more, it seems like #include can create a lot of different annoying issues. Makes me want modules a lot more haha.

engine/scene/sceneloader.cpp
jonmgomez: Apparently it is. I even tried something like this
`std::vector<const char*> sceneNames = Config::GetValue("scenes", {""});` with const char *
but still it couldn't quite get it.

engine/scene/componentdeserializer.cpp
jonmgomez: Yeah I would definitely use it again, it is a nice pattern.
Its a great way to ensure type safety and use inheritance without some crazy casting or other typing system to know what derived type an object is at runtime.
It seems the only real cons is that is that it introduces circular dependcies. It also is a little weird to have functions like `VisitComponent` because that isn't very self documenting. It get the idea across that it is using visitor on the object, but not what it is doing to that object.

engine/scene/sceneloader.cpp
cjappl: Yep, something like that

engine/scene/sceneloader.cpp
cjappl: Yeah, I think that makes sense! Good data point, thanks for checking.

engine/scene/sceneloader.h
cjappl: Yeah modules would be great for this

engine/scene/sceneloader.cpp
jonmgomez: Cool, works great!

PR #20, opened 2024-01-02
Create font strings on intialization instead of during runtime

jonmgomez:
Prevously if a text component was changed/created it would open the font file, read it, and create the character images using stb true type. This can be costly to do over and over each frame the text is changed. 

Instead `Font` objects are created at initalization and stored within the engine. It contains most the details for each font's character including the bitmap image of the character, its offsets, and sizing data. This way when recomputing the text component, it only needs to add some offsets together to get the new text data.

Previously measured loadtimes for the 3 mains scene and their new load times are as follows
Title       100-115ms -> 1-5ms
Settings 285-300ms -> 6-8ms
Game     39-40ms     -> 0-1ms
So a massive improvement! Of course there is a slight tradeoff that the initalization of the engine will take longer to complete.

Some things to note/ignore:
A lot of this is moving computations from text.cpp to font.cpp
Changes to scene files were made to adjust to new default scaling of fonts

The main idea is that the `FontBank` will load all fonts present in the config and store them in a map where a font name maps to a new font object. Text components will look for this font object, and give it the text it wants to create a component from. The font object will give back a `FontString` along with a vector of `FontCharacters`. Using the data from these objects, each text character can be made.

Another smallish change is the use of a new namespace `image` Instead of Texture holding the methods `ConvertAlphaImageToRGBA` and `FlipImageVertically` statically, they are free functions under the namespace `image`

### DIFF
--- a/components/text.cpp
+++ b/components/text.cpp
@@ -1,26 +1,27 @@
 #include "text.h"
 
 #include "config.h"
+#include "font.h"
 #include "logger.h"
-
-#define STB_TRUETYPE_IMPLEMENTATION
-#include <stb_truetype.h>
+#include "pong.h"
 
 #include <fstream>
 
 namespace pong
 {
 
-Text::Text()
+Text::Text() :
+    mText { "" },
+    mFont { Pong::GetInstance().GetFontBank().GetDefaultFontName() },
+    mScale { 1.0f }
 {
-    Text("Default", "D:/code/pong/assets/pixeloid.ttf", 1.0f, 128);
+    RecomputeText();
 }
 
-Text::Text(const std::string& text, const std::string& path, float scale, int pixelLineHeight) :
+Text::Text(const std::string& text, const std::string& fontName, float scale) :
     mText { text },
-    mFontPath { path },
-    mScale { scale },
-    mPixelLineHeight { pixelLineHeight }
+    mFont { fontName },
+    mScale { scale }
 {
     RecomputeText();
 }
@@ -44,125 +45,33 @@ BaseComponent* Text::GetBaseComponent()
 
 void Text::RecomputeText()
 {
-    std::ifstream file(mFontPath, std::ios::binary);
-    if (!file)
+    Font* font = Pong::GetInstance().GetFontBank().GetFont(mFont);
+    if (font == nullptr)
     {
-        RealTimeLogError("Failed to open true type font file: {}", mFontPath);
+        RealTimeLogError("Failed to load font: {}", mFont);
         ASSERT(false);
         return;
     }
 
-    std::vector<unsigned char> fontData((std::istreambuf_iterator<char>(file)), std::istreambuf_iterator<char>());
+    const FontString fontString = font->GetCharacters(mText);
+    const std::vector<FontCharacter>& fontCharacters = fontString.mFontCharacters;
 
-    stbtt_fontinfo font;
-    stbtt_InitFont(&font, fontData.data(), 0);
-
-    const float pixelScale = stbtt_ScaleForPixelHeight(&font, static_cast<float>(mPixelLineHeight));
-
-    float totalTextWidth = 0.0f;
-    float totalTextHeight = 0.0f;
-    float currentLineWidth = 0.0f;
-    float currentLineHeight = 0.0f;
-
-    int unscaledAscent, unscaledDescent, unscaledLineGap;
-    stbtt_GetFontVMetrics(&font, &unscaledAscent, &unscaledDescent, &unscaledLineGap);
-
-    const float ascent  = unscaledAscent  * pixelScale;
-    const float descent = unscaledDescent * pixelScale;
-    const float lineGap = unscaledLineGap * pixelScale;
-
-    float currentX = 0;
-    float currentY = 0;
-
-    for (int i = 0; i < mText.length(); i++)
+    for (const FontCharacter& character : fontCharacters)
     {
-        const char character = mText[i];
-
-        int unscaledGlpyhWidth = 0;
-	    int unscaledLeftSideBearing = 0;
-        stbtt_GetCodepointHMetrics(&font, character, &unscaledGlpyhWidth, &unscaledLeftSideBearing);
-
-        const float glpyhWidth = unscaledGlpyhWidth * pixelScale * mScale;
-        const float leftSideBearing = unscaledLeftSideBearing * pixelScale * mScale;
-
-        int xCoord1 = 0;
-        int yCoord1 = 0;
-        int xCoord2 = 0;
-        int yCoord2 = 0;
-        stbtt_GetCodepointBitmapBox(&font, character, pixelScale, pixelScale, &xCoord1, &yCoord1, &xCoord2, &yCoord2);
-
-        const int charWidth  = xCoord2 - xCoord1;
-        const int charHeight = yCoord2 - yCoord1;
-
-        auto alphaTexture = std::vector<unsigned char>();
-        alphaTexture.resize(charWidth * charHeight);
-
-        if (character == '\n')
-        {
-            currentY -= (ascent - descent + lineGap) * mScale;
-            currentX = 0;
-
-            if (currentLineWidth > totalTextWidth)
-            {
-                totalTextWidth = currentLineWidth;
-            }
-            totalTextHeight = fabs(currentY);
-            currentLineWidth = 0;
-            currentLineHeight = 0;
-        }
-        else
-        {
-            const float spaceAboveCharPixels = ascent + yCoord1;
-            const float characterYOffset = currentY - spaceAboveCharPixels * mScale;
-
-            const float quadScreenWidth  = charWidth * mScale;
-            const float quadScreenHeight = charHeight * mScale;
-
-            const float lineHeight = fabs(characterYOffset - quadScreenHeight);
-            if (lineHeight > currentLineHeight)
-            {
-                currentLineHeight = lineHeight;
-            }
-            currentLineWidth = quadScreenWidth;
-
-            // Since this uses a single character for each texture, stride is just the pixel width of the character
-            const int kStride = charWidth;
-            stbtt_MakeCodepointBitmap(&font, alphaTexture.data(), charWidth, charHeight, kStride, pixelScale, pixelScale, character);
-
-            if (i > 0)
-            {
-                currentX += leftSideBearing;
-            }
-
-            // Since the current offset is the top left, offset it to center of quad
-            const float finalXOffset = currentX + quadScreenWidth / 2.0f;
-            const float finalYOffset = characterYOffset - quadScreenHeight / 2.0f;
-
-            mCharacters.emplace_back(alphaTexture, quadScreenWidth, quadScreenHeight,
-                                     glm::vec3(finalXOffset, finalYOffset, 0.0f),
-                                     charWidth, charHeight);
-
-            currentX += glpyhWidth;
-
-            // TODO: Needs to also be checked after a \n character
-            if (i == mText.length() - 1)
-            {
-                currentX -= glpyhWidth - quadScreenWidth;
-            }
-        }
-
-        const int kern = stbtt_GetCodepointKernAdvance(&font, character, mText[i + 1]);
-        currentX += fabs(kern * mScale * pixelScale);
-        currentLineWidth = currentX;
+        const float quadWidth = character.mBitmapWidth * mScale;
+        const float quadHeight = character.mBitmapHeight * mScale;
+        const float xOffset = character.mXOffset * mScale;
+        const float yOffset = character.mYOffset * mScale;
+        mCharacters.emplace_back(character.mBitmap, quadWidth, quadHeight,
+                                 glm::vec3(xOffset, -yOffset, 0.0f),
+                                 character.mBitmapWidth, character.mBitmapHeight);
     }
 
-    if (currentLineWidth > totalTextWidth)
-    {
-        totalTextWidth = currentLineWidth;
-    }
-    totalTextHeight += currentLineHeight;
+    const float totalTextWidth = fontString.mWidth * mScale;
+    const float totalTextHeight = fontString.mHeight * mScale;
 
-    // Characters have been created with their top left corner at the origin, so offset them to the center
+    // Characters have been created with their top left corner at the origin,
+    // so offset them to the center of the whole text
     const glm::vec3 center = glm::vec3(totalTextWidth / 2.0f, -totalTextHeight / 2.0f, 0.0f);
     for (auto& character : mCharacters)
     {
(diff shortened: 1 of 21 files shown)


components/text.cpp
```diff
@@ -1,26 +1,27 @@
...
 
 #include <fstream>
 
 namespace pong
 {
 
-Text::Text()
+Text::Text() :
+    mText { "" },
+    mFont { Pong::GetInstance().GetFontBank().GetDefaultFontName() },
```
cjappl: It may make more sense to inject this dependency instead of having it be explicit. This being here means the `Text` class is basically un-unit-testable without a lot of overhead, as opposed to explicitly taking the font name as an argument.

components/text.cpp
```diff
@@ -44,125 +45,33 @@ BaseComponent* Text::GetBaseComponent()
 
 void Text::RecomputeText()
 {
-    std::ifstream file(mFontPath, std::ios::binary);
-    if (!file)
+    Font* font = Pong::GetInstance().GetFontBank().GetFont(mFont);
```
cjappl: Same RE: dependency injection

engine/graphics/font.cpp
```diff
@@ -0,0 +1,171 @@
+#include "font.h"
+
+#include "image.h"
+#include "logger.h"
+#include "utils.h"
+
+#define STB_TRUETYPE_IMPLEMENTATION
```
cjappl, line 8: Should minimally have an #ifndef STB_TRUETYPE_IMPLEMENTATION  around it, just so it doesn't blow over some previous definition of this. 


You could also consider defining this in CMake target_compile_definitions, not sure which is better or worse, I decided to put this in CMake for the rtlogger, but not sure it matters too much. Just want you to see the multiple ways of doing it.

engine/graphics/font.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include <stb_truetype.h>
+
+#include <fstream>
+
+namespace pong
+{
+
+// The font will load these characters by default.
+// If a character is not in this list, it will be loaded the first time it is used.
+static const std::string DEFAULT_FONT_CHARACTERS =
```
cjappl: good opportunity for constexpr (i think std::string supports it)

engine/graphics/font.cpp
```diff
@@ -0,0 +1,171 @@
...
+    "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!@#$%^&*()-=_+[]{}\\|;:'\",<.>/?`~ ";
+
+static constexpr int PIXEL_LINE_HEIGHT = 256;
+
+bool Font::LoadFont(const std::string& fontName, const std::string& fontPath)
```
cjappl, line 29: Alright, so the big question for me is "Can this be done in `constexpr`"? 

You claim there is a massive speedup in the font rendering times (awesome!!) but it offloads some of this work to a one time setup cost, plus probably some working memory overhead of holding it around.

It seems feasible to me that there could be some way to do this all at compile time. Essentially what you're doing is grabbing a file, reading it in, then converting it to a datatype "font". It seems entirely possible that this could happen at COMPILE TIME, which would increase your compile time, and your binary size, but reduce ram usage and startup time.

I think you may run into some third party or system dependencies that block you, but I suspect it may be worth a trial

engine/graphics/font.cpp
```diff
@@ -0,0 +1,171 @@
...
+    int unscaledLeftSideBearing = 0;
+    stbtt_GetCodepointHMetrics(&mFontInfo, character, &unscaledGlpyhWidth, &unscaledLeftSideBearing);
+
+    fontCharacter.mGlyphWidth = unscaledGlpyhWidth * mPixelScale;
+    fontCharacter.mLeftSideBearing = unscaledLeftSideBearing * mPixelScale;
+
+    int xCoord1 = 0;
+    int yCoord1 = 0;
+    int xCoord2 = 0;
+    int yCoord2 = 0;
```
cjappl, line 74: ah, brings me back to spatial.......................... good ol' C return by param

engine/graphics/font.cpp
```diff
@@ -0,0 +1,171 @@
...
+    stbtt_GetCodepointBitmapBox(&mFontInfo, character, mPixelScale, mPixelScale,
+                                &xCoord1, &yCoord1, &xCoord2, &yCoord2);
+
+    const int charWidth  = xCoord2 - xCoord1;
+    const int charHeight = yCoord2 - yCoord1;
+
+    std::vector<unsigned char> alphaTexture { 0 };
```
cjappl: Something to be very mindful of is that std::vector has a bunch of dumb initializers, two of which are very easy to mess up:

```
std::vector<int> b{3}; // creates a 1-element vector holding {3}
std::vector<int> d(3); // creates a 3-element vector holding {0, 0, 0}
```

If your intention was to create a size 1 vector with the only element being initialized to zero, you succeeded.

(just pointing this out as a frequent bug that bites every now and again, it's constantly called out in C++ books and was probably a design mistake)

engine/graphics/font.cpp
```diff
@@ -0,0 +1,171 @@
...
+            // The kerning is the amount of space between a specific character pair
+            // This can only be calculated if the next character is known
+            const int kern = stbtt_GetCodepointKernAdvance(&mFontInfo, character, text[i + 1]);
+            currentX += std::fabs(kern * mPixelScale);
+        }
+
+        fontCharacters.push_back(fontCharacter);
+    }
+
+    return { fontCharacters, totalTextWidth, totalTextHeight };
```
cjappl, line 164: Designated initializers are going to make this so much nicer in the future:

https://en.cppreference.com/w/cpp/language/aggregate_initialization

```
return { .chars = fontCharacters, .width = totalTextWidth, .height = totalTextHeight };
```

so much more explicit and less chance for bugs. C++20!

engine/graphics/fontbank.cpp
```diff
@@ -0,0 +1,87 @@
...
+        else
+        {
+            LogWarning("Default font not found: {}, using {} as default", mDefaultFontName, newDefault);
+            ASSERT(false);
+        }
+        mDefaultFontName = newDefault;
+    }
+}
+
+Font* FontBank::GetFont(const std::string& fontName)
```
cjappl, line 70: another good use for `optional` in the future

engine/graphics/image.cpp
```diff
@@ -0,0 +1,56 @@
...
+    std::vector<unsigned char> convertedImage {};
+    convertedImage.reserve(imageData.size() * 4);
+
+    const unsigned char kFullColor = 255;
+    for (auto& alpha : imageData)
+    {
+        convertedImage.emplace_back(kFullColor);
+        convertedImage.emplace_back(kFullColor);
+        convertedImage.emplace_back(kFullColor);
+        convertedImage.emplace_back(alpha);
```
cjappl: Not sure if it makes any difference here, but for simple objects (like a byte of unsigned char) `push_back` and `emplace_back` do the same thing. Not sure if there is a performance cost to `emplace`

engine/graphics/image.cpp
```diff
@@ -0,0 +1,56 @@
...
+            // For each pixel, copy all the components
+            const int pixelIndex = rowIndex + (colNumber * comp);
+            for (int componentIndex = 0; componentIndex < comp; componentIndex++)
+            {
+                flippedImage.emplace_back(imageData[pixelIndex + componentIndex]);
+            }
+        }
+    }
+
+    return flippedImage;
```
cjappl, line 60: What a beautiful opportunity for a unit test. With all the wacky index-foo you got going here it is very ripe for an off-by-one error messing your day up.

engine/graphics/image.cpp
```diff
@@ -0,0 +1,56 @@
+#include "image.h"
+
+#include "utils.h"
+
+namespace pong
```
cjappl: I believe since 17 you can nest namespaces more naturally.

This would become 

```
namespace pong::image 
{
...
} 
```

engine/graphics/image.h
```diff
@@ -0,0 +1,15 @@
...
+#include <vector>
+
+namespace pong
+{
+
+namespace image
+{
+
+std::vector<unsigned char> ConvertAlphaImageToRGBA(const std::vector<unsigned char>& imageData);
+std::vector<unsigned char> FlipImageVertically(const std::vector<unsigned char>& imageData, int width, int height, int comp);
```
cjappl: is it worth it to have an Image class? Seems like a lot of these params are internal hidden details of what an image is. By not encapsulating it, you make the code harder to change later.

render/texture.cpp
```diff
@@ -31,12 +32,9 @@ Texture::Texture(const std::string& filePath)
     }
 }
 
-Texture::Texture(const std::vector<unsigned char>& alphaImage, int width, int height)
+Texture::Texture(const std::vector<unsigned char>& rgbaImage, int width, int height)
```
cjappl, line 35: same RE: image class

cjappl (approved):
Nice.

Want to call out a good practice of you measuring the problem before attempting to fix it, and coming back with data supporting that you met your goal.

Often times if you go with what you THINK is slow instead of profiling things, you come back empty handed and sad. Always measure->change->measure.

These tasks also lead to good resume bullet points "Reduced rendering time by 33%", as it's quantified and real.

components/text.cpp
jonmgomez: Yeah I see what you mean. I guess I need to understand how realistically testable these classes are to begin with. These text objects will all require a font to already be made in order to do anything useful. But the font already has a lot of various calculations. Is it strange to have a unit test that does a lot of file i/o? Should there be an example .ttf file that is used specfically for unit testing with fonts?

engine/graphics/font.cpp
jonmgomez: So apparently std::string is not able to be constexpr. C++20 is able to, but with some caveats. However, std::string_view is able to. And actually reading a bit on string_view, I think it makes more sense for this kind of usage anyway.

engine/graphics/font.cpp
jonmgomez: Well I guess the first issue is being able to read in a file at compile time. Im not finding a real easy way to do that, and especially not since the true type font files are not plain text either. Is there a feasable way to real in a file like that at compile time?

engine/graphics/image.cpp
jonmgomez: Not a confident answer I have but it seems like the compiler will probably be smart enough to optimize it away, but `push_back` might even be a bit safer and more readable.
https://stackoverflow.com/questions/48158286/emplace-back-vs-push-back-for-primitive-types

engine/graphics/image.cpp
jonmgomez: So along with introduced the Image class I put the `pong::image` namespace with those functions, but the `Image` class under just `pong`  I think it is a bit nicer to be able to create an `Image` without doing something like `image::Image` but but at that point maybe those functions should just be static functions like before.  That would at least make more sense than being under the `Texture` class previously. Any thoughts?

engine/graphics/image.h
jonmgomez: Yeah that is a good change!

engine/graphics/font.cpp
jonmgomez: Ahh very cool! I had no idea.

engine/graphics/image.cpp
cjappl: > think it is a bit nicer to be able to create an Image without doing something like image::Image but but at that point maybe those functions should just be static functions like before.

Authors choice. I have seen many times something like `pong::image::Image`. Often that's where you get things like `using pong::image::Image` which allows you to just use it as `Image` in the file where that is called, kind of the best of both worlds.

engine/graphics/image.cpp
cjappl: Cool! Today I learned.

Seems like either is fine

engine/graphics/font.cpp
cjappl: Cool!

components/text.cpp
cjappl: Yeah I think that makes sense to have a .ttf to test with specifically.

Note that dependency injection has a bunch of benefits on top of unit testing. It makes the dependency explicit (so future readers know on instantiation, it also forces you to reckon with classes that depend on too many other things because the param list gets nasty). On top of that, you can configure different fonts easily at compile or run time.

Lastly, if you write to an interface, instead of this concrete implementation, you can quickly write new `Font` subclasses really easily without messing with additional code. That's probably overkill for now, but it's good to think about.

All of this boils down to making this class and all supporting classes easier to change in the future. Been listening to a podcast where one of the things that stuck out to me was the quote "every time you change something in your code, note and notice why it was hard to change, how could you have avoided that from the beginning?"

engine/graphics/font.cpp
cjappl: Hmm, great question and an obvious first roadblock.

I discussed the matter with Dr. ChatGPT and he came up with an interesting solution:
1. Write a python script converting the file to raw bytes
2. Output that to a header file 
3. Wrap all of that in CMake

Seems like a feasible solution, but much more of a pain in the ass than it's worth right now. Your users are just going to have to suffer 1ms startup time lag, shameful.

engine/graphics/image.cpp
```diff
@@ -2,55 +2,62 @@
...
+    convertedImage.mWidth = image.mWidth;
+    convertedImage.mHeight = image.mHeight;
+    convertedImage.mComponents = 4;
+
     return convertedImage;
 }
 
 /* Flips an image so top left element becomes bottom left element. */
-std::vector<unsigned char> FlipImageVertically(const std::vector<unsigned char>& imageData, int width, int height, int comp)
+Image FlipImageVertically(const Image& image)
```
cjappl, line 33: Much better. you see how this removed complexity and made the code easier to change in the future? now if you support 3d images (for some reason) you don't have to edit the interface. If you change how width is calculated, you can do that internally to the Image class and no one is the wiser

cjappl (approved):

components/text.cpp
jonmgomez: Cool youve convinced me!

engine/graphics/font.cpp
jonmgomez: Hmm, yeah that is a interesting way to do it. Im curious if that is very widely used.

engine/graphics/image.cpp
jonmgomez: Yeah thats a good point. Kind of forgot about `using` making that not an issue haha

components/text.h
```diff
@@ -15,26 +16,25 @@ namespace pong
...
     void SetText(const std::string& text);
-    void SetFont(const std::string& fontName);
+    void SetFont(Font* font);
     void SetScale(float scale);
     void SetColor(GLRGBAColor color);
     void RecomputeText();
 
     std::string mText {};
-    std::string mFont {};
+    Font* mFont { nullptr };
```
cjappl, line 33: I think you could get away with using a reference here and avoid the case where something is nullptr.

engine/graphics/font.cpp
cjappl: Seems a little esoteric. I think you'd have to be seriously stressed about some overhead for the user

PR #21, opened 2024-01-07
Try to remove some singletons

jonmgomez:
Keeping this as a draft. I wanted to remove some of the static global singletons from the code so that way it is better testable. This is just one way I wanted to experiment doing this.

### DIFF
--- a/assets/gameover.scene
+++ b/assets/gameover.scene
@@ -26,13 +26,13 @@
         },
         {
           "type": "Button",
-          "width": 400,
-          "height": 200
+          "width": 500,
+          "height": 250
         },
         {
           "type": "Text",
-          "text": "Play Again",
-          "scale": 0.3,
+          "text": "Play\nAgain",
+          "scale": 0.45,
           "color": {
             "r": 0,
             "g": 0,
@@ -52,13 +52,13 @@
         },
         {
           "type": "Button",
-          "width": 400,
-          "height": 200
+          "width": 500,
+          "height": 250
         },
         {
           "type": "Text",
-          "text": "Main Menu",
-          "scale": 0.3,
+          "text": "Main\nMenu",
+          "scale": 0.45,
           "color": {
             "r": 0,
             "g": 0,
@@ -78,13 +78,13 @@
         },
         {
           "type": "Button",
-          "width": 400,
-          "height": 200
+          "width": 500,
+          "height": 250
         },
         {
           "type": "Text",
           "text": "Quit",
-          "scale": 0.5,
+          "scale": 0.6,
           "color": {
             "r": 0,
             "g": 0,
--- a/assets/settings.scene
+++ b/assets/settings.scene
@@ -79,7 +79,7 @@
           "type": "Slider",
           "width": 1000.0,
           "height": 100.0,
-          "min": 0,
+          "min": 1,
           "max": 4000,
           "step": 1,
           "start_value": 0
--- a/components/behavior.cpp
+++ b/components/behavior.cpp
@@ -6,19 +6,4 @@
 namespace pong
 {
 
-void Behavior::SetTimeout(const std::chrono::duration<double>& timeout, const std::function<void()>& callback)
-{
-    Pong::GetInstance().GetTimer().AddTimer(GetGameObjectId(), timeout, callback);
-}
-
-void Behavior::PlaySound(const Sound& sound)
-{
-    Pong::GetInstance().GetAudioMixer().PlaySound(sound);
-}
-
-void Behavior::PlaySound(const Sound& sound, const glm::vec3& position)
-{
-    Pong::GetInstance().GetAudioMixer().PlaySound(sound, position);
-}
-
 } // namespace pong
--- a/components/behavior.h
+++ b/components/behavior.h
@@ -27,10 +27,6 @@ public:
     virtual void OnCollisionStay(GameObject& /*other*/) {};
     virtual void OnCollisionStop(GameObject& /*other*/) {};
 
-    void SetTimeout(const std::chrono::duration<double>& timeout, const std::function<void()>& callback);
-    void PlaySound(const Sound& sound);
-    void PlaySound(const Sound& sound, const glm::vec3& position);
-
     bool mOnStartCalled { false };
 };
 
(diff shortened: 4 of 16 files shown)


jonmgomez:
Added you here @cjappl. I was curious what you thought of doing global variables like this instead. I wanted to remove some of the singletons that way I can test things easier. So instead of accessing the timer from `Pong::GetInstance().GetTimer().SetTImeout()` is it a free function accessed under the namespace `timer::SetTimeout()` which uses the gTimer variable set to whatever the Pong instance has currently.

This way testing the timer, and future classes like Input can be tested normally by creating stack variables instead of relying on an entire Pong instance to be active. I think I do like this better, but wasn't sure if there were some other implications of doing like this. Or a easier way in general to accomplish this.

cjappl:
Cool! It's good to try some different stuff and see what the pros and cons are. As always there is no perfect solution, and there are pros and cons of everything.

If you do go with this approach:
1. I would minimally hide things behind getters and setters.

I don't think that any `gVariables` should just be accessed without a getter and a setter
You should try to avoid any public or protected `mVariables`

The reason for this is the same reasons we have talked about a lot of the changes in the past, hiding things behind getters and setters make things easier to change in the future. It encapsulates the information so if you need to access it in a different way in the future, you can change it without anyone else knowing.


2. I think this is better than the singleton approach

The main thing I like about this approach is that it explicitly instantiates things (one of the downsides of singleton). The singleton just kind of sweeping it's instantiation under the rug can lead to some pain and suffering in some cases, and explicitly setting them up and tearing them down is preferable.



As an alternative:
1. I think the optimal solution may be moving a lot of this to explicit dependencies to allow you to evaluate the architecture more easily.

Take this with a grain of salt, you should compare and contrast to other game engines and way the trade offs. 

As we discussed in the previous PR, there are a lot of benefits to making the dependencies extremely explicit, and injected. You should think about your architecture, and ensure that everything knows about JUST ENOUGH to do it's job, and no more. 

For instance. In this case there is a very sneaky hidden dependency in that `Ball` needs to have `gAudioMixer` set up before it can operate properly. If I was a developer joining this project, how would I be aware of that? It isn't a parameter that is passed in, there is no indication that there is a dependency until it crashes at some point.

You can make the dependency explicit by passing it in as a variable. At that point at least everything is on the surface.

From there, you can trivially map out your architecture. You can think "Is it reasonable that `Ball` depends on `gAudioMixer`?" The answer may be yes or no but at least you can evaluate it. You may be surprised that some dependencies you have don't make sense.

To make a concrete example of this: What if you had a collision manager, that depended on Ball and gAudioMixer (it seems reasonable that the things managing collision would need to know about the audio system). You could then strip out the OnCollisionStart method of your objects and move that behavior to the collision manager. `Ball` knowing about audio, collisions, moving through space is a lot of info to pack in one class.

The question to ask in some of these cases is "Is this class adhering to the Single Responsibility Principle" aka, "Does it have just one job?". In the existing code, the `Ball` is both a circular graphics primitive moving through space, but also handles collisions. That's at least two jobs.

The more classes that adhere to this principle, the easier future change is.

jonmgomez:
> You can make the dependency explicit by passing it in as a variable. At that point at least everything is on the surface.

Would it be feasable to do this? The main reason being that since I want to simulate a "true game engine" the best I can, classes like the `Ball` would need to inherit something that already exists within engine, that being `Behavior`. I don't think there is a good way to pass in `gAudioMixer` explicitly without setting up a special constructor for `Ball` somewhere in engine code? Im sure there is a more structured way to achieve it like inheriting from `AudioManager` gives access to `gAudioMixer`. But that also might be a bit too much setup for my liking. I did look at what Unreal Engine is trying to do:
```
#include "Kismet/GameplayStatics.h"

if (HitSound != nullptr)
{
    UGameplayStatics::PlaySoundAtLocation(this, HitSound, GetActorLocation());
}
```

cjappl:
> like inheriting from AudioManager gives access to gAudioMixer. But that also might be a bit too much setup for my liking. I did look at what Unreal Engine is trying to do:

Yeah, I don't like inheriting from AudioManager, that doesn't seem correct.

I think this example proves the prior art exists to make these static and global in some way. Can you dig into the implementation and tell how they set up this object `UGameplayStatics` and how it ensures things are initialized etc? That would be an interesting case study.

jonmgomez:
Yeah so it is using a similar global extern variable that I found in Engine.h
```
/** Global engine pointer. Can be 0 so don't use without checking. */
extern ENGINE_API class UEngine*			GEngine;
```
Which is used in GameplayStatics under static functions
```
void UGameplayStatics::PlaySoundAtLocation(const UObject* WorldContextObject, class USoundBase* Sound, FVector Location, ...)
{
    UWorld* ThisWorld = GEngine->GetWorldFromContextObject(WorldContextObject, EGetWorldErrorMode::LogAndReturnNull);
    ...
    if (FAudioDeviceHandle AudioDevice = ThisWorld->GetAudioDevice())
    {
	...
        AudioDevice->PlaySoundAtLocation(Sound, ThisWorld, ...);
    }
```
That seems to be properly intialized in these functions `FEngineLoop::PreInitPostStartupScreen()` or `FEngineLoop::Init()`
```
GEngine = GEditor = NewObject<UEditorEngine>(GetTransientPackage(), EditorEngineClass);
...
UE_LOG(LogInit, Log, TEXT("Initializing Editor Engine..."));
GEditor->InitEditor(this);
UE_LOG(LogInit, Log, TEXT("Initializing Editor Engine Completed"));
```
And that does seem to be active for the lifetime of the engine/editor itself as in UnrealEngine.cpp
```
//
// Exit the engine.
//
void UEngine::FinishDestroy()
{
        ...
	if ( !HasAnyFlags(RF_ClassDefaultObject) )
	{
		// shut down all subsystems.
		GEngine = NULL;
        }
        ...
}
```
There are only a few spots where this global variable is set, so I assume they try to guarentee it is active for the entire lifetime of the engine. Generally it seems the usage of this variable is done without nullptr checks either, although that comment suggests otherwise.

cjappl:
Cool, in that case I totally defer to the way they do things. I think given the fact that you're building a game engine, you should defer to what they do. 

This has the advantage of giving you something to look at when in doubt, and also gives you an excuse to dig into the weeds to see how things work under the hood.

jonmgomez:
Sweet, will close this for now then. And make a new pr later on fully doing this.

PR #22, opened 2024-01-10
Remove most singletons from engine creation.

jonmgomez:
A little hard to review since a lot of this is removing static keywords, removing GetInstance() calls, and removing static functions. The main idea is that in order to remove singletons, areas of the code that need to be accessed use free functions under a specific namespace. Ex to change the window's title from a user defined `Behavior`, `application::SetWindowTItle("My Game");` is used. 

Some of the singletons being removed:
Engine, Renderer, Input, Pong, ApplicationWindow

And importantly this uses depency injection to build the thing the game engine needs. The Engine class remains the top level class holding everything else `Renderer`, `ApplicationWindow`, `Input`, `Pong`, `Input` does depend on both the renderer and application window to properly use callbacks so they are passed into the constructor.

### DIFF
--- a/engine/applicationwindow.cpp
+++ b/engine/applicationwindow.cpp
@@ -37,7 +37,7 @@ void ApplicationWindow::Init()
         ASSERT(false);
     }
 
-    glfwSetFramebufferSizeCallback(mWindow, application::WindowResizeCallback);
+    glfwSetFramebufferSizeCallback(mWindow, globals::application::WindowResizeCallback);
     glfwMakeContextCurrent(mWindow);
     SetVSync(Config::GetValue("vsync", false));
 
@@ -115,7 +115,7 @@ GLFWwindow* ApplicationWindow::GetWindow() const
 
 } // namespace pong
 
-namespace pong::application
+namespace pong::globals::application
 {
 
 ApplicationWindow* gApplicationWindow { nullptr };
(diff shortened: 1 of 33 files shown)


engine/applicationwindow.cpp
```diff
@@ -120,3 +114,55 @@ GLFWwindow* ApplicationWindow::GetWindow() const
...
+}
+
+void SetWindowTitle(const std::string& title)
+{
+    GetWindowInstance()->SetWindowTitle(title);
+}
+
+void SetWindowIcon(const std::string& iconPath)
+{
+    GetWindowInstance()->SetWindowIcon(iconPath);
```
cjappl, line 155: One of the things I liked about the Unreal version of this was it made it clear it was operating on some global statics. Can you incorporate that here somehow? Via an object or a namespace or similar?

```
GlobalStatics::WindowInstance()->
```

Makes it super painfully obvious at the call site you're mucking with globals.

cjappl (approved):
Looks good to me. Just briefly skimmed though, as there is a lot there.

Comment somewhere specific if you want eyes, otherwise ship it.

engine/applicationwindow.cpp
jonmgomez: Hmm yeah what about something like
`globals::audio::PlaySound()` and `globals::application::SetWindowIcon()` under a `globals` namespace? I think that makes sense.

cjappl: yeah, just something indicating there is something going on in the globals when it happens. I think that when it's necessary to mutate global state it should be hyper obvious

